PR #319, opened 2024-11-15
Add Spanish locale

PwtKCL:
See commits for details.


src/localisation/en.json
```diff
@@ -97,7 +97,7 @@
         "errorUserDefinedFuncMsg":"User-defined function\nthis function definition contains errors",
         "errorUserDefinedVarMsg":"User-defined variable\none of the assigments for this variable contains errors",
         "EOFError": "This frame cannot be empty.",
-        "GDriveSaveFailed":"Strype coud not saved to Google Drive. Your latest changes may have not been saved, and the sync has been turned off.",
+        "GDriveSaveFailed":"Strype coud not save to Google Drive. Your latest changes may have not been saved, and the sync has been turned off.",
```
neilccbrown, line 100: One more correction: coud->could.  And actually, two lines up: assigments->assignments.  Maybe I should have a pass through at some point.

src/localisation/es.json
```diff
@@ -0,0 +1,201 @@
...
+        "resetProject":"Nuevo proyecto",
+        "resetProjectTooltip":"Restablecer el editor con el proyecto predeterminado de Strype",
+        "prefs":"Preferencias",
+        "lang":"Idioma: ",
+        "version":"versión",
+        "homepage":"Página de inicio de Strype"
+    },
+
+    "autoCompletion": {
+        "myFunctions":"Mis funciones ",
```
neilccbrown, line 183: We should probably trim the trailing space.

neilccbrown (changes requested):
Did you see that Github Actions tells us warnings that occurred during the tests?  Interesting!  Although actually we may want to disable that one because we do have a console since it's all client-side, and all those console.log look deliberate.

Anyway, just a couple of small typo-like things to fix.

src/localisation/en.json
PwtKCL: If that's the logs I'm thinking about, yes I was aware of them (but unfortunately I've noticed them only after doing the merge request of the related code). So I'll keep it the related code as is for now but will change it when doing work on bug fixes pretty soon.

src/localisation/es.json
PwtKCL: Well spotted

PwtKCL:
Regarding the warnings: I've noticed some warnings during the compilation which I left for the time being, as I need to make some code fixes later anyway so I can do it later. 
But after looking in more details I think you refer to the warnings about "no-console" ? We can discuss it when we meet.

src/localisation/en.json
PwtKCL: Done, good point and I've tried to spot other typos in a quick pass with Word (to see the spell checker indicators).

Issue #243, opened 2024-07-25
Try frames without joint child (except/finally) error not handled properly


neilccbrown:
In GitLab by @PwtKCL on Jul 25, 2024, 12:20

In Python, a try construct can't be by itself.
We don't check explicitly that a Try frame has joint children; TigerPython doesn't pick up this error, and Skulpt execution fails without giving a clear nature of the error (it gives an error message, but it's not clear).

We need to think what is best to do to handle this situation: pre-construct the Try frame? Always show error when no joint is attached? Change in TigerPython?

(remember to check if disabling a frame hides the error)

PR #322, opened 2024-11-24
Handles "try" structures wrong syntax (missing "except" or "finally")

PwtKCL:
This PR mainly addresses issue #243, and it relies on the support of a new TigerPython version (1.0.1) which handles the detection of problematic "try" structures.

However, supporting the error at parsing means that we also now need to display errors in Strype at the frame level rather than just slot, so the PR also handles that.

The PR also contains support for using available matching locales in TigerPython (v1.0.2), and rebase the npm package to the upstream version in our Strype package.json.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,7 @@
                 "sortablejs": "^1.15.0",
                 "splitpanes": "^2.4.1",
                 "textarea-caret": "^3.1.0",
-                "tigerpython-parser": "github:neilccbrown/TigerPython-Parser#fix-empty-global",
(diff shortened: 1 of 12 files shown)


PwtKCL:
> Code all seems fine but the tests have failed because it does not find `TPyParser.getLanguages()`. Are we waiting on a PR to expose this method, or do we need an "as any" cast in there?

The problem comes from not getting the right package. After your suggestion I've added a new commit that removes the TigerPython related entries in package-lock.json to force the right package version to be used and therefore the tests to succeed.

PR #323, opened 2024-11-26
Update docs

PwtKCL:
I've updated the README with 
- target of (external) links to "_blank", GitHub very stupidly ignores them but if it ever supports links like that it's better to open them in a new tab by default
- the test command seems to miss the "run" keyword, I'm not entirely sure if depending on the version of NPM or the IDE being used, that keyword can be omitted or not though (but for me, it is required)

Also, I've added a documentation guidelines file. I've thought for quite some time today about it. My document was initially written with Word. I wanted to make a README file as Neil suggested but I couldn't get the styling output I wanted to have. However, having Word isn't ideal I think. I wanted to share a PDF file but I wanted also users to be able to be able to send pull requests on the document itself, and PDF isn't great for that. Ideally, I think we would like to get HTML here in case we want to simply bind the content to our website and make it very simple to work with (for us and for whoever wants to make a change on it, I mean)? But for the moment, I suggest we leave it like that and see the best option later (I expect some fixes or changes to be made after feedback..)


neilccbrown:
I think you're right that Word isn't ideal, both for changing and also for reviewing changes in future.  Which is the section that you had trouble formatting?  I think we should be able to format the JSON parts by telling Github it's JSON (see https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks if you follow the link to https://github.com/github-linguist/linguist/blob/main/lib/linguist/languages.yml it supports Javascript and JSON).  For the yellow parts maybe we can use their alerts formatting instead: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts  The images should be doable if you commit them too alongside the README and link to them.

PwtKCL:
After discussion in our meeting, I changed the Word document to a Readme file (with the alerting formats suggested by Neil, and images instead of text for the text I couldn't show in colour). I've removed a few things from the file as suggested by Michael.
Also, added contributors to the project as we discussed with Neil.

README.md
```diff
@@ -61,6 +61,21 @@ Contributing
...
+
+Other contributors:
+- Babis Kyfonidis
+- Zak Singh
+- Aleksandr Voronstov
+
+Translators:
+- Greek: Babis Kyfonidis
+- French: Pierre Weill-Tessier (King's College London)
+- Spanish: Emma Dodoo and Nolgie Oquendo-Colon (University of Michigan)
```
neilccbrown, line 77: I think we probably don't need the institution in the translators part, just the names.

docs/translation/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+>[!NOTE]
+>Currently, all translations are taking effect in Strype when a new version of the application is released by the Strype team. So even if a translation has been received by the Strype team and included into the source code repository, it will be visible in the online tool only when the Strype team releases a new version.
+
+## How to retrieve or create the translation files
+
+To retrieve the translation files, you will need to locate them inside the Strype source code repository hosted in GitHub (<a href="https://github.com/k-pet-group/Strype" target="_blank">https://github.com/k-pet-group/Strype</a>).
+
+The translation files for all supported locales are inside the `/src/localisation/` folder, which contains **one subfolder per locale (as a 2-letter code)**. For example, the translations for English are in <code>/src/localisation/**en**/</code>, those of Spanish are in <code>/src/localisation/**es**/</code> and so on.
```
neilccbrown, line 24: Might be worth linking to this page https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes which lists the two letter codes (in case they don't know the code for the new language).

docs/translation/README.md
```diff
@@ -0,0 +1,91 @@
...
+>[!CAUTION]
+>Only the **text** values need to be translated.</span>
+
+## How to do the translations
+
+The language reference for all translations is the English locale. Therefore, the reference files are the files located in the `/en/` subfolder (see sections [1](#how-to-retrieve-or-create-the-translation-files) and [2](#how-to-understand-the-translations-files) for details about finding the files and understanding their formats).
+
+**Each** *key:value* pair of a translation locale file **need to match those of the equivalent English locale file**, and as explained above, **only** the text *values* need to be translated.
+
+Here are a few rules to follow when doing the translations of the text *values*.
```
neilccbrown, line 56: Probably worth mentioning to save the files in UTF-8 (but can point out this is the default for most modern text editors).

neilccbrown (commented):
Just a couple of small comments but otherwise fine to merge.

docs/translation/README.md
PwtKCL: Good idea.

PwtKCL:
Addressed Neil's suggestions in the latest commit -- I merge now.

Issue #257, opened 2024-07-26
Dragging frames while holding option (Mac) or ctrl (Windows) should copy the frame


neilccbrown:
The drag should be modifiable by holding a key to turn it into a copy.  The usual behaviour is that the key can be pressed after the drag begins (and releasing it turns back into a normal drag), and a "+" or similar is shown somewhere to indicate that it will be a copy rather than a move.

PR #324, opened 2024-11-27
New implementation of drag & drop (continued)

PwtKCL:
In this PR, I've fixed a bug in the new implementation of the drag & drop that could cause frames to be moved in an illegal position.
It also contains the feature request of  #257

### DIFF
--- a/src/components/Frame.vue
+++ b/src/components/Frame.vue
@@ -466,7 +466,7 @@ export default Vue.extend({
             // Not all frames should be duplicated (e.g. Else)
             // The target id, for a duplication, should be the same as the copied frame 
             // except if that frame has joint frames: the target is the last joint frame.
-            const targetFrameJointFrames = this.appStore.getJointFramesForFrameId(this.frameId, "all");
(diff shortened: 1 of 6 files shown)


PwtKCL:
There are a few more things I will need to add, so I close that PR for now.

neilccbrown:
Looks fine to merge, so I will do so.  What's the replacement for the draggable group logic?  I was using that in the new OOP stuff but I guess I need to make a change in a different part of the code now (e.g. class frames should only allow comments and functions)?

PwtKCL:
> Looks fine to merge, so I will do so. What's the replacement for the draggable group logic? I was using that in the new OOP stuff but I guess I need to make a change in a different part of the code now (e.g. class frames should only allow comments and functions)?

There is no replacement as such. The groups were used as one layer for the checks of drag and drop - the other layer is a programmatic check against what frames are allowed at a given position. We can check together where in the code if needed;
Regarding the draggable group, there is no replacement because 1) the second layer of checking suffices and 2) if you see the commits of that PR, we seemed to only use the "all" group in the method I had deleted (...!).

Issue #252, opened 2024-07-25
Paste frames over a selection should replace over the selection


neilccbrown:
In GitLab by @PwtKCL on Jul 25, 2024, 15:16

Check Stride

PR #328, opened 2024-11-29
Paste frames over a frame selection

PwtKCL:
This PR implements one of the functionality listed for 1.0.1 (#252).

While working on that functionality, I realised I missed one case for handling the listing of allowed frames at a given position when we had a selection at the moment of pasting (regardless it was overwriting the frames or not), and we had copied or cut the frames from the context menu. The list of frames was the wrapping frames instead of the frames allowed at that position regardless the selection.

I've added a change of timeout value in this PR, as suggested by Neil to see if that helps passing the tests.

### DIFF
--- a/src/components/CaretContainer.vue
+++ b/src/components/CaretContainer.vue
@@ -270,7 +270,7 @@ export default Vue.extend({
 
         doPaste(skipDisableCheck?: boolean) : void {
             let pasteDestination: CurrentFrame = {id: this.frameId, caretPosition: this.caretAssignedPosition};            
-            const stateBeforeChanges = JSON.parse(JSON.stringify(this.appStore.$state));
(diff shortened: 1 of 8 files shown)


src/components/CaretContainer.vue
```diff
@@ -266,26 +267,48 @@ export default Vue.extend({
                 this.doPaste();
             }
         },
-        
+
         doPaste(skipDisableCheck?: boolean) : void {
+            let pasteDestination: CurrentFrame = {id: this.frameId, caretPosition: this.caretAssignedPosition};            
+            const stateBeforeChanges = JSON.parse(JSON.stringify(this.appStore.$state));
```
neilccbrown, line 273: I think we can use lodash's cloneDeep for this but I don't have a strong preference.

neilccbrown (commented):
Just one comment to consider, otherwise all fine.

PR #329, opened 2024-11-29
Upload videos and screenshots to Github on test failure

neilccbrown:
As discussed yesterday, this makes the tests upload the screenshots and videos of the test failure (in a zip) to Github when it fails the tests -- but still fails in this case.  I also removed a test which I think is flaky because of Cypress sometimes having an odd bug and sometimes not.  Now, at least on the latest run of this branch, the tests all pass.


PwtKCL:
All good - that will be helpful !!! thanks

PR #330, opened 2024-12-01
Create German translation

mkolling:
created folder and main json file


PwtKCL (commented):
The folder and file name are fine, but since the translation is not there (the file is the English version) should I wait for another iteration?

mkolling (commented):
file translated now

PR #334, opened 2024-12-02
Frame shortcuts and list updates

PwtKCL:


neilccbrown:
All fine.  The test that is failing is failing for a reason: it was for checking that Ctrl-C inserts a comment on "c" if you release ctrl before pressing c.  That doesn't apply directly now that "c" doesn't insert a comment, but it would be good to reinstate the test when we add classes.  I suggest you comment out basics.cy.ts from line 376 to the end, and I'll uncomment and adjust it in my oop branch that I'm working on.

PR #336, opened 2024-12-02
Add support for classes

neilccbrown:
This is a bit bigger than I had thought it was, but I think that's mainly because I renamed function definitions -> definitions which caused a lot of changes.  If it's easier, I could merge that one commit first, then do the rest in a second PR.

The basic principle is simple: we add a class frame which can only contain function frames or comment frames.  The main complication is that we want to automatically add "self" as the first (uneditable) parameter on functions inside classes, but not have it on functions outside classes.  Plus, we want to add a comma when the user is entering/has entered their own params, but not show the comma otherwise.  (Although, handily, Python actually allows the trailing comma anyway, which is useful for code generation because we don't need to test if there are any params following.)

I also put back that test you just commented out for ctrl-c with a selection.


PwtKCL:
There is another change that can be made. In store.ts, look for `getFuncDefsFrameContainerId` which can also be renamed (and therefore all its calls too).

I have also made a suggestion for simplification and ,if that is a valid point, a stricter check that can be done -- but otherwise it's all good.

PwtKCL (commented):
Sorry.. I had made the comments yesterday but didn't validate the review!

neilccbrown:
All suggested changes made.

PR #340, opened 2024-12-02
Localisation cleanup

PwtKCL:
See commits for details.


PwtKCL:
@mkolling I don't know if just by mentioning your profile you'll get updates on this pull request progress. Otherwise, check on the right hand side panel of this page, you will see a notification bell symbol (either clear or stroked) that allows you to subscribe to notifications if not already set by default.

Issue #311, opened 2024-10-29
Opening a Python file adds and saves over cloud storage


neilccbrown:
If you have a project open that is saving to cloud storage, and you use "Open" to open a .py file, the resulting opened file is saved to the same cloud storage place, thus overwriting the project.  In fact, it also retains the previous code, it just adds the new code from the .py file alongside.  This seems surprising to me from an "Open" action; I think it should remove all existing code, then open the .py file, and not save over the cloud storage.

PwtKCL:
Should blank the current code AND disconnect from Google Drive.

Issue #310, opened 2024-10-25
Warn when resuming work on cloud project


neilccbrown:
Here's a sequence of steps which I perform reasonably often:
 - I work on a project in Strype
 - I save this project to Google Drive; now all my changes will be synced to Google Drive
 - I work on the project
 - I close the browser window
 - I go back to Strype

At this point I'm pleased because my previous project is there, so I resume editing.  But if I was a user, I think I wouldn't realise that my project is actually (a) not loaded from Google Drive (if I made changes on another machine in the mean-time) and (b) is no longer saving to Google Drive.

I think we should inform the user with a modal dialog:

"Last time you were editing this project, you were saving to Google Drive.  Do you want to reconnect to Google Drive in order to synchronise your changes?  If you do not, you will not have the latest version from Google Drive and your changes will not be saved back there."

Options are:
 - Yes, sign me in
 - No, continue without

PR #345, opened 2024-12-03
A few cloud-related updates (for 1.0.1)

PwtKCL:
This PR includes changes that are planned for 1.0.1:
- resync Google Drive projects on browser (re)load #310 
- clean up project on Python file opening #311 

It also contains code cleanup, notably to replace the configuration environment variable "isPurePython" to "isPython" (partly because it's just an easier name, partly because I stupidly used "isPython" elsewhere and didn't realise it was actually not working...


src/App.vue
```diff
@@ -401,7 +410,25 @@ export default Vue.extend({
                 this.appStore.setStateFromJSONStr( 
                     {
                         stateJSONStr: savedState,
-                        callBack: () => {},
+                        callBack: (loadedSuccesfully) => {
+                            // When a file had been reloaded and it was previously synced with Google Drive, we want to ask the user
+                            // about reloading the project from Google Drive again (if the file changed on Google Drive side)
```
neilccbrown, line 415: I think this is just an issue with the comment and not the code, but I think the bracket is not right: we always want to ask even if the project didn't change in Google Drive because it's about setting up the syncing as much it is about reloading the changed state.

neilccbrown (commented):
Code looks good, just one small comment about a comment.

src/App.vue
PwtKCL: The comment didn't mean to say we ask only if the file has changed. I wanted to say "the file could have changed". But that's true that even if that one reason, it's also just about doing the connection with Google Drive as you mentioned. I'll remove it.

src/localisation/en/en_main.json
```diff
@@ -95,7 +95,8 @@
     "gdriveLocation": "Folder:",
     "gdriveAllStrypeFiles": "All Strype files",
     "gdriveTab": "My Drive",
-    "editorErrors": "Errors"
+    "editorErrors": "Errors",
+    "resyncToGDAtStartup": "Last time you were editing this project, you were saving to Google Drive.\nDo you want to reconnect to Google Drive in order to synchronise your changes?\nIf you do not, you will not have the latest version from Google Drive and your changes will not be saved back there."
```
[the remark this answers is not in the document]
PwtKCL: Of course, I took it from your issue ticket 😆

PR #350, opened 2024-12-05
More work on canvas graphics

neilccbrown:
Not an urgent PR, I can deploy the graphics test without waiting for the merge.  This is various changes Michael suggested for the graphics API, mainly around the collision detection (which now uses the actor name everywhere, which has been renamed "tag" here).  Also a couple of fixes to make the world bounds -399, -299 to 400, 300 as just discussed.  Probably easiest to view all at once rather than by individual commit (because some things changed twice).


PwtKCL (commented):
All good.

PR #351, opened 2024-12-05
Drag and Drop improvemements

PwtKCL:
The PR contains changes for adding a greyed background in the companion image while we wait for HTML2Canvas to generate the image.
We are aware there is still an issue with lags from the library itself, but we will leave it for 1.0.1.
The PR also contains fixes with Firefox, one that is brought by the drag and drap, another that was there beforehand.

### DIFF
--- a/src/components/LabelSlot.vue
+++ b/src/components/LabelSlot.vue
@@ -9,6 +9,8 @@
             @click.stop="onGetCaret"
             @slotGotCaret="onGetCaret"
             @slotLostCaret="onLoseCaret"
+            @mouseenter="handleMouseEnterLeave(true)"
+            @mouseleave="handleMouseEnterLeave(false)"
             @keydown.up="onUDKeyDown($event)"
             @keydown.down="onUDKeyDown($event)"
             @keydown.prevent.stop.esc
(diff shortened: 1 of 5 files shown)


src/components/LabelSlot.vue
```diff
@@ -427,6 +429,20 @@ export default Vue.extend({
             }
         },
         
+        handleMouseEnterLeave(isEntering: boolean) {
+            // There is a bug with how Firefox handles editable text HTML elements contained in a draggable div.
+            // We need to detect when the mouse is entering/leaving the text element to disable/enable the div's
+            // draggable attribute. 
+            // Because the frames are imbricated, we need to do that for all the frames hierarchy up ot the frames container.
```
neilccbrown, line 436: Today I learned a new word!  "Imbricated"...

src/helpers/editor.ts
```diff
@@ -991,11 +991,14 @@ export function notifyDragStarted(frameId?: number):void {
 
     // Add companion "image" (canvas) to the cursor - we use HTML2Canvas. 
     // The element to generate an image of is either the frame passed as argument
-    // or the shadow element containing the current selection.
-    const draggingEl = document.getElementById(getFrameUID(frameId??(useStore().frameObjects[useStore().selectedFrames[0]].parentId)));
-    if(draggingEl){
-        html2canvas(draggingEl, html2canvasOptions);
-    }
+    // or the selected frame's parent which will be cropped.
+    setTimeout(() => {
```
neilccbrown, line 995: Maybe just add a comment here about why we use the setTimeout (so that the grey rectangle shows up while html2canvas does its work, I believe?)

neilccbrown (commented):
Just one small suggestion otherwise all fine to merge.

src/components/LabelSlot.vue
PwtKCL: Aahahh yes? But it's good you mention this comment, because I think it's not correct - I really wanted to say "nested" so I'll change the word in the next commit.

PR #359, opened 2024-12-11
Final (?) updates to graphics functionality

neilccbrown:
This is the last of the graphics additions/fixes that I had on my list for development, so we can then switch to prolonged testing in the new year.  It's all fairly short anyway, but the individual commits will explain what they are doing.

### DIFF
--- a/public/public_libraries/strype/graphics.py
+++ b/public/public_libraries/strype/graphics.py
@@ -854,10 +854,9 @@ def stop():
 
 _last_frame = _time.time()
 
-def pause(actions_per_second):
+def pause(actions_per_second = 25):
     """
-    Waits for a suitable amount of time since the last call to pause() to allow you to take
-    the given number of actions per second.  This is almost always used like follows:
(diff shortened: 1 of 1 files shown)


public/public_libraries/strype/graphics.py
```diff
@@ -849,3 +849,30 @@ def stop():
...
     """
     raise SystemExit()
+
+_last_frame = _time.time()
+
+def pause(actions_per_second):
+    """
+    Waits for a suitable amount of time since the last call to pause() to allow you to take
+    the given number of actions per second.  This is almost always used like follows:
+
```
PwtKCL, line 861: I have some difficulty to understand the part "the given number of actions per second" in the comment. Later you say "Where 30 is the number of times you want to do those actions per second.", which I think is clear, so did you mean to say "the given number of times per second" in the comment here?

neilccbrown: Ok, I've adjusted it (and also added a default value for the parameter, so many users may not need to think about it).

PR #367, opened 2024-12-12
Switched to TigerPython for autocomplete

neilccbrown:
As discussed already, this switches to TigerPython for autocomplete.  I've turned off the failing tests for now, with the hope we can restore them in future.  For now I'm targeting a non-main branch in the central repository, so we can choose later when we want to deploy this.


PwtKCL (approved):
So far it looks like the change also simplifies our code a lot. Seems fine.

Issue #358, opened 2024-12-10
Add plus sign to dragged frame when option-dragging


mkolling:
Add a little (+) symbol (see MacOS Finder) to the dragged frame preview when duplicating a frame by dragging (Option-drag).

PwtKCL:
As discussed in meeting, we will add the "+" symbol at the destination frame cursor rather than at the image (easier to implement).
The "+" symbol is a green disc with a clear white "+" inside, like in macOS.

(Note: we discussed having the mouse pointer changed to the CSS supported "copy" cursor, but it's not very clear to see in Chrome at least - and we also discussed having our own cursor, but if we do it for this case, we may need to do for all cursors as otherwise there will be incoherence...)

PR #368, opened 2024-12-13
Drag and drop: improve the UI cues for "duplicate + D&D" 

PwtKCL:
These improvements include having a "+" symbol at the destination frame caret (#358) and also not blurring the souce frame(s) when "duplicate" is activated during drag and drop.

Also, for drag and drop in general, I've "forced" the mouse cursor to not be "pointer" when we are exactly at the destination caret container. 
(The "pointer" cursor at the caret container in general is questionnable, I think we did that when we systematically had a space dedicated to the caret at _every_ caret positions in earlier versions of Strype. Now that the space is collapsed most of the times, maybe it's not needed. However, I've not removed it because it might still make sense to show a different mouse cursor when we are at the blue caret.)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -349,10 +349,10 @@ export default Vue.extend({
             }
 
             // Handling the notification for doing duplication with drag and drop.
-            // We don't really care if another key is hit along ctrl/meta, we only look that
+            // We don't really care if another key is hit along ctrl/option, we only look that
             // we are currently in a drag and drop action, and notify the current caret candidate for drop that
             // the action requires frame duplication.
-            if(this.appStore.isDraggingFrame && (event.ctrlKey || event.metaKey)){
+            if(this.appStore.isDraggingFrame && (event.ctrlKey || event.altKey)){
                 addDuplicateActionOnFramesDnD();
                 event.preventDefault();
                 event.stopImmediatePropagation();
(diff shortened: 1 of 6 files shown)


src/App.vue
```diff
@@ -347,6 +347,33 @@ export default Vue.extend({
             if(event.key.toLowerCase() == "contextmenu"){
                 this.appStore.isContextMenuKeyboardShortcutUsed  = true;
             }
+
+            // Handling the notification for doing duplication with drag and drop.
+            // We don't really care if another key is hit along ctrl/meta, we only look that
+            // we are currently in a drag and drop action, and notify the current caret candidate for drop that
+            // the action requires frame duplication.
+            if(this.appStore.isDraggingFrame && (event.ctrlKey || event.metaKey)){
```
neilccbrown, line 355: I always find these key names confusing, but on Mac it is "Option" to drag-copy, and that is altKey in Javascript, not metaKey.

src/App.vue
```diff
@@ -347,6 +347,33 @@ export default Vue.extend({
...
+                event.stopImmediatePropagation();
+                event.stopPropagation();
+                return;
+            }
+        });
+
+        // There are only a few cases when we need to handle key up events
+        window.addEventListener("keyup", (event) => {
+            // Handling the notification for not doing duplication anymore with drag and drop.
+            // We don't really care if another key is hit along ctrl/meta, we only look that
```
neilccbrown, line 367: Also need to adjust the comment here (and above) to say Option or Alt instead of Meta.

neilccbrown (commented):
Need to fix the key but otherwise all looks fine.

src/App.vue
PwtKCL: Very nice catch ! It's my (very stupid) mistake, even considering the key codes are rightfully used in another part of the code, I should have been more careful! Thanks

src/App.vue
PwtKCL: I've added them on the condition,  I had forgotten them. Note that "ctrlKey" an alike can't be used for "keyup" events, because if the modifier key is the only one pressed, the modifier flag on the event will be false. We need to use the key code/value instead.

PR #373, opened 2024-12-17
Switch to TigerPython for autocomplete

neilccbrown:
I've stopped working, honest...

In my other branch I've got TigerPython working with the graphics API.  This has convinced me that this branch, which just switches autocomplete to TigerPython is ready to merge into the main branch.  You've seen all the code already except for the last commit which just updates TigerPython version.  So it doesn't need review as such, other than it's up to you whether you think this is an okay time to merge it in.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "strype-editor",
-    "version": "1.0.0",
+    "version": "2024.12.12",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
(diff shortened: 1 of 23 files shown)


PwtKCL:
Thanks for doing this when it's your break ! I think it's good we have the change as soon as possible in the main branch so we can implicitly test it when doing further development. So I'll merge it in.

On a side note, would the changes made by Tobias require a bump of the TPP version? As it stands, that's still the same version.

PR #375, opened 2025-01-15
Fix few bugs

PwtKCL:
These are pretty simple fixes I've added for now. I'd like to get them sorted first, before I can work on bigger changes in a proper new branch later.

### DIFF
--- a/src/store/store.ts
+++ b/src/store/store.ts
@@ -392,6 +392,14 @@ export const useStore = defineStore("app", {
             const filteredCommands: {[id: string]: AddFrameCommandDef[]} = cloneDeep(addCommandsDefs);
             const allowedJointCommand: {[id: string]: AddFrameCommandDef[]} = {};
 
+            // If frames are selected, we only show the frames commands that can be used for wrapping.
+            // That can be done in a systematic rule below, but for the case being inside the definitions section,
+            // we will need to check things a bit differently: making sure there are only comments selected as 
+            // functions or classes cannot be wrapped around.
+            const isSelectingInsideDefsSection = ((currentFrame.id == useStore().getFuncDefsFrameContainerId || state.frameObjects[currentFrame.id].parentId == useStore().getFuncDefsFrameContainerId) 
+                && state.selectedFrames.length > 0);
+            const areSelectedFramesAllComments = state.selectedFrames.every((frameId) => state.frameObjects[frameId].frameType.type == AllFrameTypesIdentifier.comment);
+
             // for each shortcut we get a list of the corresponding commands
             for (const frameShortcut in addCommandsDefs) {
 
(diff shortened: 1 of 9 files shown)


src/store/store.ts
```diff
@@ -392,6 +392,14 @@ export const useStore = defineStore("app", {
             const filteredCommands: {[id: string]: AddFrameCommandDef[]} = cloneDeep(addCommandsDefs);
             const allowedJointCommand: {[id: string]: AddFrameCommandDef[]} = {};
 
+            // If frames are selected, we only show the frames commands that can be used for wrapping.
+            // That can be done in a systematic rule below, but for the case being inside the definitions section,
+            // we will need to check things a bit differently: making sure there are only comments selected as 
+            // functions or classes cannot be wrapped around.
```
neilccbrown, line 398: I'm not sure I mind whether comments can be wrapped.  I wonder if technically we should allow functions+comments to be wrapped into a class... but on the other hand, how often will people think to try this?  So I'm fine with any configuration: no wrapping in definitions, only comments, or only comments+functions.

neilccbrown (commented):
One comment about wrapping, but not necessarily anything to be done.  Otherwise looks fine to merge.

src/store/store.ts
PwtKCL: Unless we decided that nothing can be wrapped at all, that section of code will require update when the OOP is rolled out completely. At the moment I can't make a rule that includes the class frame as it's not yet inside this version of Strype.
It may be quite a niche functionality, but it's not impossible some users may want that, especially if they start off with a single simple Python code and want to scaffold towards making classes. 

We can discuss it in our group meeting - let's wait for it.

PwtKCL: Conclusion from the meeting: we won't support wrapping _at all_ (either with functions or classes).

src/store/store.ts
```diff
@@ -413,8 +410,7 @@ export const useStore = defineStore("app", {
                 // We need a special case for when we are directly inside the definitions section, and some frames are selected, to not allow wrapping
                 // a function, or a class, with another function or a class.
                 if(isSelectingInsideDefsSection) {
-                    filteredCommands[frameShortcut] = filteredCommands[frameShortcut].filter((x) => !forbiddenTypes.includes(x.type.type) && !x.type.isJointFrame
-                    && (lookingForTargetPos || (areSelectedFramesAllComments && x.type.allowChildren)));
+                    filteredCommands[frameShortcut] = filteredCommands[frameShortcut].filter((x) => false);
```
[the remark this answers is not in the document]
PwtKCL: I agree. I was worried a direct assignment would create some odd behaviour. It seems not so I'll change it.

PR #380, opened 2025-01-23
Add autocomplete support to the canvas graphics module

neilccbrown:
The diff for graphics.py seems to be quite screwed up; it shows me removing a bunch of code and re-adding it but really I just inserted some bits in the middle.  The overall aim of this PR is to add code completion support to the canvas-graphics branch, based on the recent switch to TigerPython autocomplete in the main branch.  We add some special type annotations in comments, then we have some code to extract these annotations and feed them to TigerPython, so for example it knows that load_image returns an EditableImage.  We only need the annotations for return types, not parameters, as we don't need those for autocomplete.  The new raw-loader library lets us load the .py files directly in Typescript.


PwtKCL:
Code-wise I think the branch can be merged. There is one test failing that seems unrelated to the new tests added, so it shouldn't be too complex to sort.
I suggest to commit a fix for the tests to pass, if possible, then the PR can be merged fine.

PR #382, opened 2025-01-23
Fixed detection of focus in LabelSlotsStructure

neilccbrown:
I need to fix this so that when focus moves in or out of an "empty" class method header, the "self" part should change to "self," if there is focus in the slot (i.e. with comma), and "self" if the focus is not there.  The existing methods weren't working right so I've made this change.... but I'm not 100% sure about it, so see what you think.  I've kept this small so that you can focus on just this change.

### DIFF
--- a/src/components/LabelSlot.vue
+++ b/src/components/LabelSlot.vue
@@ -234,11 +234,7 @@ export default Vue.extend({
         },
 
         focused(): boolean {
-            // We need to keep update of the label slots structure's "isFocused" flag, because using the keyboard to navigate will not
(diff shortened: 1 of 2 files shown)


PwtKCL:
It's strange we had something not working before - especially as a comment mentioned the case that I think you were testing. 
I think either it was and change was made that removed some logic, or some cases were just not thought thoroughly when testing.

It seems the fix is fine. Just make sure all interactions with the mouse and the keyboard results in expected outcome.
The focus/blur with contenteditable elements, and divs, is quite a mess..

Issue #360, opened 2024-12-11
Issues with opening a project stored on the file system when opening a Google Drive project with Safari


PwtKCL:
When opening a project, Strype triggers an auto-save (for webstorage and Google Drive), then load the selected file.
This auto-save happens when the user click "OK" on the open project dialog, and then when saving is done, we trigger the file explorer or Google Drive to select the project to load.

However, if the current project is from Google drive and we want to open a project on the file system, this action won't work with Safari because the trigger to open the file explorer isn't directly happening within the modal callback (considered to be the user action).

PwtKCL:
After meeting: we won't make a fix on v1.0.1, but soon after (1.0.2?).
Change to make: we want to get rid of the auto-saving (except in WebStorage). 
Therefore, opening another project when we had loaded an existing project will now trigger a confirmation to save the existing project first, then open the other project.
We also want to show "saved" and "modified" status on the command pane, last time it had been saved, and a button for saving.
The UI aspects can be discussed again later.

PR #386, opened 2025-02-04
Changes of the saving mechanisms (and a few other fixes)

### LINKED ISSUE #385: Click "outside" the Menu should close the menu (in particular the green empty space under the code)

PwtKCL:
The bulk of this PR concerns a change in our project sync/saving mechanisms (see #360):
- removes auto-save via syncing with Google Drive
- ask if changes in an edited code should be discarded/saved when opening another project (this applies to any situation: modified projects that are on the browser, or initially retrieved from Google Drive or the file system)
- add a saved/modified state for all projects with a "last saved" date (for browser projects, "saved" and no specific date are shown)

**For this bulk parts, I still need to amend the message in the discard/save popup and change the "reconnect to Drive" mechanism when open Strype and detect a saved Google Drive project** so ignore these in this PR .

Other small fixes:
- cursor showing for drag and drop, notably ctrl+DnD (which didn't show in empty frame containers -- **still need to check why Safari doesn't work**
- close Strype menu when other parts of the UI are clicked (fixes #385)


neilccbrown:
Looks like the failing checks all say the same: "AssertionError: Timed out retrying after 4000ms: Expected to find content: 'This device' within the element: <button.btn.btn-primary> but never did.".  They are trying to load a .py file by clicking "This device" then "OK", you'll need to update how that works in the paste-python.cy.ts file (lines 104-105)

src/components/Commands.vue
```diff
@@ -3,9 +3,12 @@
         <div :class="{'no-PEA-commands': true, 'cropped': isExpandedPEA}" @wheel.stop>
             <div class="project-name-container">
                 <span class="project-name">{{projectName}}</span>
-                <div v-if="isSyncingInGoogleDrive" :title="autoSaveGDriveTooltip">
-                    <img :src="require('@/assets/images/logoGDrive.png')" alt="Google Drive" class="gdrive-logo"/>   
-                    <span class="gdrive-sync-label" v-if="isSyncingInGoogleDrive && !isEditorContentModifiedFlag" v-t="'appMessage.autosaveGDrive'" />
+                <div @mouseover="getLastProjectSavedDateTooltip" :title="lastProjectSavedDateTooltip">
+                    <img v-if="isProjectFromGoogleDrive" :src="require('@/assets/images/logoGDrive.png')" alt="Google Drive" class="project-target-logo"/> 
+                    <img v-else-if="isProjectFromFS" :src="require('@/assets/images/FSicon.png')" alt="Google Drive" class="project-target-logo"/> 
+                    <img v-else :src="require('@/assets/images/empty.png')" alt="Google Drive" class="project-target-logo"/>
```
neilccbrown, line 9: I think the alt text needs adjusting on this line and the line above.

src/components/Commands.vue
```diff
@@ -575,6 +579,63 @@ export default Vue.extend({
             }          
         },
 
+        getLastProjectSavedDateTooltip() {
+            // We show an indication about the last saved date of the document.
+            // There shouldn't be a case when we get a date that is no set, but to prevent weird behaviour
```
neilccbrown, line 584: I feel like this is maybe meant to refer to a specific bit of code, but I'm not sure.

src/components/Commands.vue
```diff
@@ -575,6 +579,63 @@ export default Vue.extend({
...
+                    : this.$i18n.t("appMessage.lastSavedOn1Hour") as string;
+            }
+            else if(lastSaveDateTickDiff > 60000){
+                // Less than an hour but more than a minute ago (60 s * 1000 ms)
+                const nbMins = lastSaveDateTickDiff / 60000;
+                toolTipVal = (nbMins > 1)
+                    ? this.$i18n.t("appMessage.lastSavedOnNMins", {nbLastSave: Math.round(nbMins)}) as string
+                    : this.$i18n.t("appMessage.lastSavedOn1Min") as string;
+
+            }
```
neilccbrown, line 628: I don't really mind since the code is written, but I imagine there is a library for this kind of thing.

src/components/GoogleDrive.vue
```diff
@@ -405,41 +411,38 @@ export default Vue.extend({
...
+                params: {fields: "name, modifiedTime"},
+            }).execute((resp) => {
+                if(this.currentAction == "loadAsResync"){
                     fileName = resp.name;
-                });
-            }
+                }
+                // The date conversion works fine because Google Drive API uses RFC 3339 date format
+                lastSaveDate = Date.parse(resp.modifiedTime);
+            });
```
neilccbrown, line 445: Is this code synchronous?  i.e. can be we sure this request finishes before we start the next one?

neilccbrown (commented):
A few comments to look at but otherwise all seems fine.

src/components/Commands.vue
PwtKCL: Hail to the copy/paste... thanks for spotting this !

src/components/Commands.vue
PwtKCL: That was to be with another comment a bit lower (I didn't get the right comment portion when I moved bits of code around. I move it to its correct position now.

src/components/Commands.vue
PwtKCL: Yes maybe. Let's keep it though, it might just be easier to write our own rather than look up for something that really does exactly what we want (with the localised labels). 
We can always update it later.

src/components/GoogleDrive.vue
PwtKCL: I think so, there is nothing about it indicating it's a promise.

PR #387, opened 2025-02-12
Add support for image literals

neilccbrown:
(In future this will be expanded to media literals, i.e. including sound, but for now it just handles images.)

This adds image literals into Strype, as discussed.  There's a new MediaSlot which basically occupies a slot entry, and appears as a small inline image.  The main complications are around getting cursor behaviour and deleting working correctly.  Notably:
 - media literals need to be considered width 1 even though they don't have any text, because otherwise you can't distinguish the cursor position immediately before them and immediately after them.
 - media literals need to have a blank operator immediately before and after them, and always have a plain text slot before/after that, like brackets do, for the same reason: you need to be able to put your cursor directly next to the image to be able to add text, or to be able to delete the image literal.

I've added some unit tests for some of these behaviours but it will probably need a bit more human testing before release (as part of the general graphics test).

I know there's a TODO left in the test file; that will be part of the next PR, alongside some other bits (e.g. previewing the image on hover).


src/store/store.ts
```diff
@@ -883,6 +883,31 @@ export const useStore = defineStore("app", {
...
+                // We are adding a media item. In this context, the function arguments have this meaning:
+                // operatorOrBracket: the mediaType
+                // lhsCode: the code on the slot that precedes the structured slot [resp. string slot] we insert
+                // rhsCode: the code on the slot that follows the structured slot [resp. string slot] we insert
+                // midCode: if provided, the code that should be added as a base field within the structured slot [the string quote] we insert 
+                //          this makes sense when we highlight a text and wrap it with brackets [resp. quotes]
+                // Adding a new bracketed slot means we also add the empty operators "around" it:
+                // we replace the slot where we add the brackets into this:
+                // <base slot (LHS)><empty operator><bracket [resp. string] slot><empty operator><basic slot (RHS)>
+
```
PwtKCL, line 896: The "[resp. xxx]" bits can be removed in the comment above, because we are in the specific case of images only.

tests/cypress/e2e/media-literals.cy.ts
```diff
@@ -161,4 +161,37 @@ describe("Paste image literals", () => {
...
+    it("Can delete pasted image with backspace", () => {
+        cy.readFile("public/graphics_images/cat-test.jpg", null).then((catJPEG) => {
+            focusEditorPasteAndClear();
+            enterImports();
+            cy.get("body").type(" set_background(");
+            cy.wait(1000);
+            (cy.focused() as any).paste(catJPEG, "image/jpeg");
+            cy.wait(1000);
+            // A yellowy-greeny colour from his eye:
+            cy.get("body").type("{backspace}\"red");
```
PwtKCL, line 174: tiny detail, I suppose that should be something like "a plain red background" ?

tests/cypress/e2e/media-literals.cy.ts
```diff
@@ -161,4 +161,37 @@ describe("Paste image literals", () => {
...
+    it("Can delete pasted image with delete", () => {
+        cy.readFile("public/graphics_images/cat-test.jpg", null).then((catJPEG) => {
+            focusEditorPasteAndClear();
+            enterImports();
+            cy.get("body").type(" set_background(");
+            cy.wait(1000);
+            (cy.focused() as any).paste(catJPEG, "image/jpeg");
+            cy.wait(1000);
+            // A yellowy-greeny colour from his eye:
+            cy.get("body").type("{leftarrow}{del}\"red");
```
PwtKCL, line 189: and here too?

src/helpers/editor.ts
```diff
@@ -239,13 +239,15 @@ export function getFrameLabelSlotLiteralCodeAndFocus(frameLabelStruct: HTMLEleme
...
     frameLabelStruct.querySelectorAll(".labelSlot-input,.labelSlot-image").forEach((spanElement) => {
-        if (spanElement instanceof HTMLImageElement) {
+        if (spanElement.classList.contains("labelSlot-image")) {
             const code = spanElement.getAttribute("data-code");
             // We add the code, but also record the image literal for later manipulation:
             if (code) {
                 uiLiteralCode += code;
                 imageLiterals.push({code: code, mediaType: spanElement.getAttribute("data-mediaType") ?? ""});
             }
+            // Media literals are considered to be one character wise:
```
PwtKCL, line 249: small typo: "wide"

src/helpers/editor.ts
```diff
@@ -1424,11 +1426,35 @@ export const parseCodeLiteral = (codeLiteral: string, flags?: {isInsideString?:
...
+                        resStructSlot.operators.splice(j, 0, {code: ""});
+                        resStructSlot.fields.splice(j, 0, {code: before} as BaseSlot);
+                        // Need to compensate for what we just added:
+                        j += 1;
+                    }
+                    if (after.length > 0 || j == resStructSlot.fields.length || !isFieldBaseSlot(resStructSlot.fields[j+1]) || resStructSlot.operators[j].code != "") {
+                        // If there's no plain slot after or the operator isn't blank, we need to add a blank operator and blank field:
+                        resStructSlot.operators.splice(j, 0, {code: ""});
+                        resStructSlot.fields.splice(j+1, 0, {code: after} as BaseSlot);
+                    }
```
PwtKCL, line 1457: If I get it correct: in the "before" case you need to compensate (`j+=1`) so you don't read the image again when we get to the next item of the main loop.
For the the after case, I see it's not useful because you will still be reading past the image position - but since we know we've added something that won't contain an image (i.e. the blank fields) should we just skip it anyway?

tests/cypress/e2e/media-literals.cy.ts
```diff
@@ -213,6 +213,30 @@ describe("Paste image literals", () => {
             checkGraphicsCanvasContent("paste-and-color-pick-with-delete");
         });
     });
+
+    it("Can paste image in existing slot", () => {
+        cy.readFile("public/graphics_images/cat-test.jpg", null).then((catJPEG) => {
```
PwtKCL, line 218: We talk about it last meeting but I didn't really pay attention: why would it make sense to paste an image in the middle of some text? Maybe it's fine to paste at the beginning or the end of a slot with text, but in the middle I'm wondering if we should?

PwtKCL (approved):
There are few some comments related fixes I think can be done. 
Some other parts are mainly about discussion but code-wise it all looks great so merging is fine.
It's a very exciting feature :)

neilccbrown:
Thanks -- I've added another commit to adjust all the comments as suggested, then will merge.

PR #391, opened 2025-02-17
Add support for sound literals

neilccbrown:
I still have bugs to fix in the handling of media literals (as showed up in last meeting) but I'd already made a start on sound literal support so I thought I'd finish and merge that, then go back to look at the bugs.

It's relatively straightforward: media literals can now also be sound literals, and we add support for loading sounds from a base64 "data:...." URL.  The image representation for the sound is drawn with the classic green on black waveform.  We store this in a cache that expires after a while (in case the user adds a sound then deletes it).  It looks like this in the editor:

image


PwtKCL (approved):
I wonder if we should have a sort of fallback image for sounds. It seems unlikely the sound representation fails, but if it does, we may want to have a static image that is not like a blank sound. (I guess we can check if something went wrong at places like if we get a 0 size buffer or wherever else that indicates something didn't work fine.) 

Some other comments that are not impacting this PR directly:

- Regarding microbit: that’s a very good call to disable the media slots for this version of Strype. Thinking of it, maybe we could have it still but behave a bit differently (like showing the LEDs drawing) and having the appropriate methods ? We can always add this as a future feature.
- For the sound representation, I’m not entirely sure it visually look great (but the idea to show the sound representation in itself is pretty cool). Most people won’t make much of the sound representation and I think it’s quite expected to have a basic image like a megaphone or speaker for a sound. Let’s see later when we’ll see the demo during our Thursday meeting.

**The PR can be merged now though**

PR #392, opened 2025-02-17
Change behaviour when a project, that was previously saved / opened with Google Drive, is reloaded in the browser

PwtKCL:
As per the title say: we used to propose users to load the current version from Google Drive in the browser when reloading a project that was previously retrieved/saved with Google Drive;
We decided to change this behaviour to propose to _save_ it instead.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -72,7 +72,7 @@
             <span v-t="'appMessage.editorFileUploadWrongVersion'" />                
         </ModalDlg>
         <ModalDlg :dlgId="resyncGDAtStartupModalDlgId" :useYesNo="true" :okCustomTitle="$t('buttonLabel.yesSign')" :cancelCustomTitle="$t('buttonLabel.noContinueWithout')">
-            <span style="white-space:pre-wrap">{{ $t('appMessage.resyncToGDAtStartup') }}</span>
+            <span style="white-space:pre-wrap" v-html="$t('appMessage.resyncToGDAtStartup')"></span>
         </ModalDlg>
         <div :id="getSkulptBackendTurtleDivId" class="hidden"></div>
         <canvas v-show="appStore.isDraggingFrame" :id="getCompanionDndCanvasId" class="companion-canvas-dnd"/>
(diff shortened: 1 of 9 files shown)


src/App.vue
```diff
@@ -72,7 +72,7 @@
             <span v-t="'appMessage.editorFileUploadWrongVersion'" />                
         </ModalDlg>
         <ModalDlg :dlgId="resyncGDAtStartupModalDlgId" :useYesNo="true" :okCustomTitle="$t('buttonLabel.yesSign')" :cancelCustomTitle="$t('buttonLabel.noContinueWithout')">
-            <span style="white-space:pre-wrap">{{ $t('appMessage.resyncToGDAtStartup') }}</span>
+            <span style="white-space:pre-wrap" v-html="$t('appMessage.resyncToGDAtStartup')"></span>
```
neilccbrown, line 75: Out of interest, what is the difference with the change here?

src/localisation/en/en_main.json
```diff
@@ -110,7 +111,7 @@
     "gdriveAllStrypeFiles": "All Strype files",
     "gdriveTab": "My Drive",
     "editorErrors": "Errors",
-    "resyncToGDAtStartup": "Last time you were editing this project, you were saving to Google Drive.\nDo you want to reconnect to Google Drive in order to synchronise your changes?\nIf you do not, you will not have the latest version from Google Drive and your changes will not be saved back there."
+    "resyncToGDAtStartup": "<b>Reconnect and <u>save</u> to Google Drive?</b>\n\nLast time you were editing, you were saving to Google Drive.\nIf you want to <u>load</u> changes from Google Drive, click 'Do not connect' and use 'Open' from the menu."
```
neilccbrown, line 114: I think maybe add the word instead in the second sentence just to help make clear it's an alternative. ("If you want to instead load changes...")

neilccbrown (commented):
One small comment otherwise fine to merge.

src/App.vue
PwtKCL: It has to do with formatting. If the text in the locale file contains some HTML tags, like "<b>" it needs to be put in the html attribute to be interpreted as such. Putting the text directly inside the text node of the element will render those HTML tags literally rather than interpreting them as HTML.

PR #395, opened 2025-02-19
Added the ability to put assignment statements into the definitions

neilccbrown:
Very short PR!  I think this is all that's needed, though?  This is going into the OOP branch, which has already turned the function definitions section into "definitions"


PwtKCL:
Seems that's enough yes, as the rules for handling frames nesting logic relies on that part of the code.

Issue #374, opened 2024-12-24
Caret can disappear after deleting an operator


neilccbrown:
Steps to reproduce, in Firefox in main release:

 - Create an if statement.
 - Type an expression like "a >= biggest"
 - Delete the right-hand side with backspace, then backspace once more to delete the "=" (this is the crucial step, I think)
 - Type a right-hand side (like "biggest" again)
 - Press enter

The frame cursor should show inside the if, but it does not.

PR #398, opened 2025-02-20
Double operator deletion issue

PwtKCL:
This PR fixes the bug reported in #374 when deleting part of a double operator would keep the UI in a caret-less state.
The other commit (about the drag and drop) is Neil's fix that has been checked in a previous PR and therefore can be ignore here.


neilccbrown:
Looks fine, will just wait for checks to complete before merging.

PR #401, opened 2025-03-02
Added support for pasting (or loading Python files) that contain classes

neilccbrown:
I realised that the OOP branch supports adding classes in the editor, but not parsing them from Python (when you paste or load a .py file), so this PR fixes that.

### DIFF
--- a/src/helpers/pythonToFrames.ts
+++ b/src/helpers/pythonToFrames.ts
@@ -612,8 +612,51 @@ function copyFramesFromPython(p: ParsedConcreteTree, s : CopyState) : CopyState 
         break;
     case Sk.ParseTables.sym.funcdef:
         // First child is keyword, second is the name, third is params, fourth is colon, fifth is body
-        s = makeAndAddFrameWithBody(p, AllFrameTypesIdentifier.funcdef, [1, 2], 4, s);
+        s = makeAndAddFrameWithBody(p, AllFrameTypesIdentifier.funcdef, [1, 2], 4, s, (f) => {
+            if (s.parent?.frameType.type == AllFrameTypesIdentifier.classdef) {
+                // We remove the first param from the start of function params,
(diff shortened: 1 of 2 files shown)


src/helpers/pythonToFrames.ts
```diff
@@ -612,8 +612,51 @@ function copyFramesFromPython(p: ParsedConcreteTree, s : CopyState) : CopyState
         break;
     case Sk.ParseTables.sym.funcdef:
         // First child is keyword, second is the name, third is params, fourth is colon, fifth is body
-        s = makeAndAddFrameWithBody(p, AllFrameTypesIdentifier.funcdef, [1, 2], 4, s);
+        s = makeAndAddFrameWithBody(p, AllFrameTypesIdentifier.funcdef, [1, 2], 4, s, (f) => {
+            if (s.parent?.frameType.type == AllFrameTypesIdentifier.classdef) {
+                // We remove the first param from the start of function params,
+                // assuming it is the self parameter that we add automatically.
+                const params = f.labelSlotsDict[1];
```
PwtKCL, line 619: I think we should probably still check the value of the first parameters. There is no guarantee the Python file is sound.

PwtKCL (commented):
I have one small comment regarding removing "self". Otherwise, all good.

src/helpers/pythonToFrames.ts
neilccbrown: Technically, Python allows you to name the first parameter anything you like, but it still refers to self because its first, so we remove it whatever it's called (we do check there is actually 1+ parameters before trying to remove).  Does that answer the question?

PwtKCL: Yes and no. I didn't think of the case the "self" is named differently - I was thinking if the code just ignored it. So then there is nothing we can do but what you've already done...

PR #402, opened 2025-03-02
Add a dialog to support editing sound literals

neilccbrown:
This adds EditSoundDlg, which is the sound equivalent of EditImageDlg.  It's very similar in structure; we use the same image cropper to provide a UI to allow trimming the sound, and we provide a slider to allow changing the volume.  There's a button to play/stop a preview of the sound, and a button to normalise volume.  Looks like this:

image

Also fixed up the UI on buttons in the media preview popup dialog by using b-button instead of button and laying it out slightly differently.


PwtKCL (approved):
All good. I will make a PR for the French matching translations (this time, a PR not a direct commit as this will now be merged)

PR #405, opened 2025-03-13
Adding Python as allowed files Strype can load

### LINKED ISSUE #302: Also allow Python files to be loaded from Google Drive
In GitLab by @PwtKCL on Sep 24, 2024, 12:01

PwtKCL:
This PR fixes #302 to allow loading Python files from Google Drive.
Therefore, the changes include support for getting Python files at loading from Google Drive, and also make change in the Google Drive Picker to have a Python related tab as such:
image


src/components/GoogleDrive.vue
```diff
@@ -540,69 +540,93 @@ export default Vue.extend({
...
-                    // (because the error message would have been shown before for normal opening from the Drive picker)
-                    if(this.openSharedProjectFileId.length > 0){
-                        (this.$root.$children[0] as InstanceType<typeof App>).finaliseOpenShareProject("errorMessage.retrievedSharedGenericProject", reason);
-                    }
-                });
+                }
+            },
+            // Case of errors
+            (resp) => {
+                if(resp.status == 404){
```
neilccbrown, line 624: Did you mean to handle 404 specifically or did you want to show this for all errors?  What will the user see if there is another error code (e.g. 403, 500)?

neilccbrown (commented):
Generally fine, just one comment to look at.

src/components/GoogleDrive.vue
PwtKCL: I wanted to show it for 404 in particular which is what would be returned when a resource doesn't exist. But you're right, we should probably cover the larger case of any errors.

Issue #349, opened 2024-12-05
Strype menu update


PwtKCL:
Add an "example" section in the menu with a few examples of Strype code.

We also want to reorganise the menu, something like
Open
Save
Save as
< sep>
new 
< sep>
_Example_
  example1
  example2
  example3
  . . . 
< sep>
_Preferences_
  <language>

neilccbrown:
Examples were added in PR #474

PR #407, opened 2025-03-14
Share project functionality has been modified + menu

PwtKCL:
There are 2 bits in the PR : moving the sharing functionality in the Strype menu, and change the Strype menu.

For sharing projects, the mechanism is very similar (user-wise) but the link is now only copied to the clipboard when the "Copy link" button is hit. (That is for the issue with Safari requiring a user gesture, as Neil rightfully mentioned in the previous PR, **it needs to be tested on macOS**). 
There is also now a **concurrent timer involved**. The timing values are set _to my rough guess_ but I'm happy to adjust to what seems more reasonable... The timer is used to avoid a pending modal dialog that never gets the "Copy link" button enabled. 
Also, there is a delay for the first time the dialog is opened to get the link to avoid, in case of an error, a too fast action if the user didn't have time to read or decide what to do...
Finally, the keyboard shortcut for sharing a project **needs to be test on macOS** (SHIFT-CMD-L). If it doesn't work we will drop the shortcut altogether.

For the menu, changes are for the order of the elements and their presentation. The new menu looks like this for the standard editor version:
image
and like that for the microbit version:
image


src/components/Menu.vue
```diff
@@ -575,14 +671,113 @@ export default Vue.extend({
...
+                        if(createPermissionSucceeded){
+                            // We have set the file public on Google Drive, now we retrieve the sharing link.
+                            gapi.client.request({
+                                path: "https://www.googleapis.com/drive/v3/files/" + gdVueComponent.saveFileId,
+                                method: "GET",
+                                params: {fields: "webViewLink"},
+                            })
+                                .then((resp) => {
+                                    // We got the link or not, but we can only make it useful or show an error *if the user is still expecting this sharing mode from the dialog (if not, we just return)
+                                    if(this.areShareProjectActionStillValid(forShareMode)){
```
neilccbrown, line 720: What happens if the user shows the dialog, selects the non-public option, then the link arrives from Google Drive, and then the user selects the public option?  I'm not sure I fully follow the code here, but will that link get discarded rather than being stored in case they change the selected option back to public?

neilccbrown (commented):
Looks fine but think about the comment on the control flow.  I feel like we should be storing the link result regardless of the UI selection.  I.e. we should always start generating the public link, always store it, and show it once ready.  The UI just determines which one we copy to the clipboard once we click OK.  But I might misunderstand the control flow.

src/components/Menu.vue
PwtKCL: It should work fine, if I did it correctly...
 
I'll try to explain it the best I can by writing...

The logic is like that (saying you start from when Strype.org is opened):
the popup shows with "public" share, we wait 2 seconds then generate the public link
  --> if the user doesn't change the type of sharing, the public link is generated (control by  `if(initialCall && !this.shareProjectInitialCall){}`)
  ---> if the user changes the type of sharing, we trigger a new Drive link generation
    --> if the initial generation (for public) returns, it won't do anything because we detect that the sharing mode has changed OR the dialog is closed already (control by `areShareProjectActionStillValid(forShareMode)`)

Each switch of the sharing mode will call for a new link generation (we disable the button again), and any previous hanging call to the link generation should result in doing nothing because when the link gets ready, or there is an error we want to display, we check the sharing mode is the right one, and the dialog is open (`areShareProjectActionStillValid(forShareMode)`)

The only case, I think,  this would **not** work is 1) opened the sharing dialog, 2) a long call to getting the link happens, 3) user closed the dialog, 4) user went to another project and reopened the share dialog and 5) by the time the user did anything the first call to getting the link in 2 returns and we are on the same sharing mode.

But I think that's likely not possible because getting the links shouldn't take very long, and there would still require time for the user to switch projects. Also, it's likely the user will close the dialog if it is taking too long (or the 10 seconds time will kick in as timeout and close the dialog).

neilccbrown: With this logic though, if the user selects non-public then public again we will ignore the result of the first call and wait for the result of the second call?  That seems a bit wasteful if we could instead just store the result of the first call regardless of the UI and then they can click back and forth without generating any further calls or needing any further wait.  i.e. I think it will work fine, but I wonder if it's a bit over-elaborate compared to just requesting the public link once and storing it when it arrives.

PwtKCL: You're right, it would generate the link again, and you're right also that's is a waste of resources.

I think that in our context, your suggestion applies well because the link for non-public sharing isn't actually requiring any call to any method that could fail or take time.
So we can just get the right sharing link (i.e. the stored public one or the non-public one that is generated on the fly) as soon as we click "copy link".

I'll amend it later in another commit.

PwtKCL:
Partially fixes #349 for the menu order.

Issue #363, opened 2024-12-12
PEA presentation on large screens


PwtKCL:
When size of te viewport allows, we may want to stack up the 2 tabs of the PEA in the command pane.

PwtKCL:
We have made the change for the having 4 different layout views.
The different views are toggled by buttons that are where the expand/collapse button was.
They are:
- collapsed and tabbed (the intial layout as before)
- collapsed and stacked,
- expanded and tabbed (the view we had with "expand" before)
- expaded and side-by-side.

We add a slider between the frame commands and the PEA.

PwtKCL:
Another change along with those mentioned above: the layout is saved _with the project_.
That means when a project is loaded, the layout changes are brought back in the UI.
Note that it will save any layout changes. For example, if the project was in its default layout (tabbed collapse PEA), and the user changed the divider position between the PEA and the add frame commands, then changed the layout to expanded PEA, at reload the expanded PEA will be showing and the divider between PEA and commands will also be remembered.

PR #409, opened 2025-03-20
New Python Execution Area layout toggles

PwtKCL:
This PR is for #363 which focuses on having the Python Execution Area (PEA) in different layouts.
I initially wanted to split the PR in smaller parts but in the end it went to a single one - sorry about this, I think it's easier for this PR to work on the code review looking at the file changes in whole rather than commit by commit.

The new layouts combine toggling the expand/collapse views (that we had before) with having the PEA "panes" tabbed or stacked. The layout toggle buttons are icons appearing in place of the expand/collapse previous toggle button, as illustrated here:
image

There is also now a new splitter between the "add frame" commands panel and the PEA. When the UI is loaded the first time, the PEA will stick to a 4:3 ratio as in previous versions. Once a splitter is moved or a layout is changed, the ratio isn't in force anymore.
However, the min size of the collapsed PEA is still the 4:3 ratio equivalent.

**Note that in future work the layout will be saved as a user setting, and at the moment the browser's window resize isn't tackled.**
Also, and I will remove some of the literals for IDs and CSS classes to have them set in variables.


src/components/PythonExecutionArea.vue
```diff
@@ -486,6 +576,11 @@ export default Vue.extend({
         border-color: lightgray !important;
     }
 
+    .pea-no-tabs-controls {
+        color: transparent;
+        padding: 8.8px 0 7.8px 0px;
```
neilccbrown, line 581: Does this need to be sub-pixel amounts?

neilccbrown (commented):
Just a couple of comments to think about.

(I thought initially: this is quite small really, then realised Github had collapsed three large file diffs...)

neilccbrown:
Oh, also: it seems all the tests are failing with the same Typescript error, so that needs looking into.

PwtKCL:
> Oh, also: it seems all the tests are failing with the same Typescript error, so that needs looking into.

I had checked the microbit version still works fine with my changes... but I only checked by looking at the webpage, not the devtool console for errors..... So the error was some use of our Python standard editor version-specific bits that were used in some places where it didn't get rendered conditionally properly.

src/components/PythonExecutionArea.vue
PwtKCL: To be honest, I simply looked for whatever came up with the style Bootstrap generates for the PEA tabs. As I wanted to keep the same size of PEA "header", I strictly used that.

public/index.html
```diff
@@ -25,12 +25,32 @@
     <strong>We're sorry but Strype doesn't work without JavaScript enabled.
         Please enable it to continue.</strong>
   </noscript>
+  
+  <div id="StrypeSVGIconsSprite" style="display: none;">
+    <!--We need to load the SVG sprite for Strype ASAP, so it can be used by the app's components, see script below. -->
```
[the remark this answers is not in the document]
PwtKCL: Yes actually, you're right... that index.html isn't containing much for the app so it can totally get the SVG content inline. Good point.

src/components/PythonExecutionArea.vue
PwtKCL: Back on having a look at changing them anyway: it seems it doesn't make a difference. Actually, in both situations, that PEA header is slight changing size when we swap between tabbed and split layout views. 
In both cases, it's almost not noticeable so I'll just choose the full pixel values as suggested by Neil.

PR #412, opened 2025-04-01
Save and restore the layout with the project

PwtKCL:
The main part of this PR is about saving the layout/styling of the Strype UI and restore it on a project (re)load.
Because of timing and denouncing, a few timeouts are added but shouldn't be too problematic.
The idea is that on project load, we reset the UI to a default state and apply the changes from there on.

Other bits of the changes are for setting new minimum panel sizes for the different splitters of the UI, as well fixing a few bugs with layout / style handling.

### DIFF
--- a/src/store/store.ts
+++ b/src/store/store.ts
@@ -2404,9 +2404,10 @@ export const useStore = defineStore("app", {
                         // for ease of coding, we register a "one time" event listener on the modal
                         const execSetStateFunction = (event: BvModalEvent, dlgId: string) => {
                             if((event.trigger == "ok" || event.trigger=="event") && dlgId == getImportDiffVersionModalDlgId()){
-                                this.doSetStateFromJSONStr(newStateStr);      
-                                vm.$root.$off("bv::modal::hide", execSetStateFunction); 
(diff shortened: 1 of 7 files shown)


src/store/store.ts
```diff
@@ -74,10 +80,29 @@ export const useStore = defineStore("app", {
 
             isEditing: false,
 
+            /* These state properties are for saving the layout of the UI.
+             * Some are initally set to UNDEFINED so it we can detect default state:
+             * (for example, the PEA will take its 4:3 ratio size when the commands/PEA 
+             * splitter has not changed, we it needs to be seen as such. )
```
neilccbrown, line 86: Small mistake in comment: extra word?

neilccbrown (commented):
Generally seems fine, although it's obviously fairly intricate!  One small comment but it's not even that essential to address, can always do it in another PR.

src/store/store.ts
PwtKCL: Thanks for pointing out - I'll change the comment and merge.

PwtKCL:
Fixes #363

PR #414, opened 2025-04-03
Save Strype locale outside the editor's state

PwtKCL:
Previously, Strype's locale was saved with the state and therefore with the project. Whenever a project was loaded, the locale it was saved with was then applied to Strype accordingly.

This change is now decoupling a project to the Strype locale and save the locale on a "session", or for a browser rather, instead.
This is how it works now:
- the language is no longer saved in the state as before, but it is set in **another state dedicated to settings**
- when Strype is loaded (with that new change), we look if the locale is saved in _the local storage of the browser_
---- if the locale was saved, we use it to update the UI's locale,
---- if the locale wasn't saved (case of first time loading Strype since the new change), we try to detect the browser's language and use it as the Strype locale (if we support it) otherwise English is set.

_Every change of the locale (programmatically or via the UI) is saved in the browser locale storage._

The other unrelated of this PR are some style fixing for Safari (not that the font change isn't too important as it will be change with Neil's changes soon), and a fix for a small issue with Googe Drive (explained in the commit's details).

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -688,8 +688,8 @@ export default Vue.extend({
             // if they want to use the detected (supported) language. 
             // If they refused or if we can't retrieve anything at all, we use English as default.
             let strypeSessionLocale = "en"; // default locale
-            let canCheckBrowserLocale = false;
-            if(typeof Storage !== undefined) {
+            let checkBrowserLocale = false;
+            if(typeof Storage !== "undefined") {
                 const savedSettingsState: typeof this.settingsStore = JSON.parse(localStorage.getItem(AutoSaveKeyNames.settingsState)??"{}");
                 if(savedSettingsState.locale) {
                     strypeSessionLocale = savedSettingsState.locale;
@@ -698,14 +698,14 @@ export default Vue.extend({
                     // There is no locale saved. Maybe the user wants to use the default English, but maybe
                     // they would like to use another language and their working environment won't save it,
                     // so we can ask them based on the browser's locale if they want to switch.
-                    canCheckBrowserLocale = true;
+                    checkBrowserLocale = true;
                 }
             }
             else{
-                canCheckBrowserLocale = true;
+                checkBrowserLocale = true;
             }
 
-            if(canCheckBrowserLocale){
+            if(checkBrowserLocale){
                 // We didn't retrieve a locale, but we can check if the browser's locale isn't English
                 // and use it for Strype if we provide that locale
                 const foundLanguange = navigator.language?.toLowerCase();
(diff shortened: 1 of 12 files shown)


src/App.vue
```diff
@@ -664,19 +668,62 @@ export default Vue.extend({
...
             }
         },
 
+        setStrypeLocale() {
+            // We need to retrieve Strype's language (session) if it exists in the localStorage.
+            // If we didn't retrieve it, we try to infer the browser's language and ask the user
+            // if they want to use the detected (supported) language. 
+            // If they refused or if we can't retrieve anything at all, we use English as default.
+            let strypeSessionLocale = "en"; // default locale
+            let canCheckBrowserLocale = false;
```
neilccbrown, line 691: I think "canCheck" is misleading here.  It's not whether we can but whether we should.  I'd go for "checkBrowserLocale" instead.

src/App.vue
```diff
@@ -1036,58 +1083,61 @@ export default Vue.extend({
...
-                useStore().showMessage(msg, 10000);
-            }
-            else {
-                // Clear the current existing code (i.e. frames) of the editor
-                this.appStore.clearAllFrames();
-
-                copyFramesFromParsedPython(s.imports.join("\n"), STRYPE_LOCATION.IMPORTS_SECTION);
-                if (useStore().copiedSelectionFrameIds.length > 0) {
-                    getCaretContainerComponent(getFrameComponent(-1) as InstanceType<typeof FrameContainer>).doPaste(true);
+        setStateFromPythonFile(completeSource: string, fileName: string, lastSaveDate: number, fileLocation?: FileSystemFileHandle) : Promise<void> {
```
neilccbrown, line 1086: It's a bit hard to tell from the diff because of the indent change, but out of interest, why did this become a promise?

neilccbrown (commented):
Just a couple of comments but otherwise looks fine.

src/App.vue
```diff
@@ -664,19 +668,62 @@ export default Vue.extend({
...
         },
 
+        setStrypeLocale() {
+            // We need to retrieve Strype's language (session) if it exists in the localStorage.
+            // If we didn't retrieve it, we try to infer the browser's language and ask the user
+            // if they want to use the detected (supported) language. 
+            // If they refused or if we can't retrieve anything at all, we use English as default.
+            let strypeSessionLocale = "en"; // default locale
+            let canCheckBrowserLocale = false;
+            if(typeof Storage !== undefined) {
```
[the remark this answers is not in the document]
PwtKCL: Very good point !!! I didn't copy the existing code properly, that's well spotted ! It should be with quotes.

PR #416, opened 2025-04-04
Change the main font and make us self-hosted

neilccbrown:
The new font is based on Atkinson Hyperlegible Next.  I've called it "AHN-Strype" to give an indication to its source, but also denote that we changed it.  The full list of changes:
 - Removed slash from zero
 - Added the left-arrow symbol that we use for assignment
 - Changed the underscore to have more space at the sides, to make it easier to identify consecutive underscores (e.g. in "__init__").
 - Enlarged the dot, comma, colon, semi-colon to make them more visible.

These are served as woff/woff2 from our servers.  I haven't removed Source Sans Pro yet, I think that might make sense as part of a future PR to perhaps remove all use of Google Fonts, and maybe FontAwesome.  (An initial search for "<i " only found five uses of FontAwesome)

You can see the new font right now on my repo's live version: https://neilccbrown.github.io/Strype/

Also moved logos into src/assets/logos as discussed.

Also self-hosted the Inconsolata font and fontawesome, as discussed.  This means Strype now runs locally even with no internet connection (I just tried).  It fails to load the Google Drive APIs but that doesn't stop Strype working.


PwtKCL (commented):
I think it's fine to merge but:
- I think yes we can remove Source Sans Pro. With your change, is it used actually anywhere now?
- We way remove the use of "Inconsalata" now if the new font works with the stretch. Or keep for another PR, that's up to you. 

Regarding fontawesome: I also looked at it after our talk yesterday and yes, it's not much used and it's mainly for the micro:bit API discovery. We could replace it with your own SVG icons later.

PwtKCL (approved):
I had an issue with seeing inconsolata font stretched on my mac VM, but maybe it's a problem on my side -- so we should monitor it.
Note: removing the SRI for the new font links seems logical since we host the resource. I don't know if we'll get some problem with future CASA validation, but either it will work fine, or we'll need to justify the SRI isn't making sense on self hosted links, or we will add SRI regardless in future.
Merging now.

Issue #269, opened 2024-08-01
Keyboard shortcuts can create frames in folded sections


neilccbrown:
Put the frame cursor somewhere in the main code.  Then collapse the main code using the mouse to click on the "My code" header.  When press a shortcut like 'i' or 't', and unfold the code; there will be a new frame added.  A similar problem occurs if focus was in text slot; it moves to frame cursor after and same thing happens) I think there's three possible fixes:

 - Don't allow frame cursor focus in folded sections (but what if they are all folded?  where is focus?).
 - Prevent shortcuts doing anything if section is folded.
 - Don't allow folding the sections in the first place (I was never convinced it was too helpful, especially when there's only three sections.)

PwtKCL:
While we need to discuss which approach to follow, we also need to consider these if folding sections stay:
- drag and drop of frames 
- errors in a folded section
- other frames actions: delete, paste etc that can add / remove frames

PR #417, opened 2025-04-07
Folding sections revamp

PwtKCL:
In answer to bug #269, we have decided to remove the expand/collapse toggle from "My code" section, and also show something in the interface when a non-empty frame container was collapsed. Both changes are illustrated here (imports are empty, definitions are not):
image

With this change I have also reduced the space around the expand/collapse icons (and set a consistent size for all browsers). 

In this PR also:
- fix a bug: ensure that when a frame is collapsed and contains the frame cursor, the frame cursor gets pushed to the next available frame container
- fix a bug with clicking near a collapsed frame container: the frame cursor would get inside, but the frame container would not expand,
- fix a bug with same problem as above but for undo/redo
- fix a bug with same problem as above but for navigating an error located inside collapsed frame container,
- fix a bug with drop positions handling during drag and drop, when a frame container was collapsed,
- add a feature for hovering long enough a collapsed frame container during D&D would expand that frame container and allow showing drop positions

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1109,16 +1109,16 @@ export default Vue.extend({
 
                     copyFramesFromParsedPython(s.imports.join("\n"), STRYPE_LOCATION.IMPORTS_SECTION);
                     if (useStore().copiedSelectionFrameIds.length > 0) {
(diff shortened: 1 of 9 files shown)


neilccbrown:
All looks fine.

PR #418, opened 2025-04-08
Add a spinner inside the share project's dialog "OK" button

PwtKCL:
Based on what we said during the meeting, this PR is for adding the spinner in the "OK" button of the share project dialog.
_This PR is based on top of the "folding section" PR: so it can be checked later after the "folding section" PR is merged, for ease of the code review._


neilccbrown:
I think Github has become a bit confused because some of the same commits were in the other PR, and I'm not totally sure what will happen if I merge.  I suggest merging the current main branch into this one then pushing again, and then I think it will show one commit needed to merge, rather than 10.  (It's fine otherwise.)

PwtKCL:
As discussed with Neil: I close the PR to redo it with the appropriate context.

PR #419, opened 2025-04-08
Input event + selection fixes

neilccbrown:
Sorry, this is big and nasty but I think it's most of the way there.  All the tests pass as it stands; I plan to do another round of testing (manual and adding some more automatic tests, e.g. for wrapping in strings) and then hopefully it's ready for everyone to test.

As before, the general plan is that we are letting the browser's natural behaviour for contenteditable occur, and then we watch afterwards for adjusting slots (e.g. operator or bracket entered) or overtyping (which then just becomes navigation).  This PR was intended to focus on selecting.  It turns out that for selection to work we had to adjust the contenteditable hierarchy: many browsers don't like the hierarchy from my first PR where I had `contenteditable > non-editable > contenteditable`, so now it's `contenteditable > contenteditable`, and I removed one div level that I don't think we needed.

What happens now when you select across slots and delete is that it all works; the browser does what you'd hope.  Our job is to (a) prevent selections where we wouldn't want to allow deletion (e.g. half in, half out of a bracket) and (b) adjust our internal state to match after.  The latter caused an issue with Vue because the user would delete, we would notice and update our intended state, but then Vue wasn't syncing the tree which caused issues (e.g. the user had deleted the hidden autocomplete elements via the browser, but Vue wasn't restorying them).  So there's this new refactorCount variable which basically forces Vue into updating.  I found a good algorithm for updating this which means we only do it when necessary, which avoids the issues I had told you about with autocomplete needing to save its state.

Also: this adds the playwright tests, due to cypress not being able to test shift-arrows for selection.  The new playwright tests run in parallel to the cypress ones which is nice, as it doesn't slow down the overall duration of the actions that way.  Oh although I need to go back and figure out scssVars properly in playwright.  We can't use the cypress trick because we are in different processes, so playwright can only transfer serialisable values across the boundary so we can't transfer the item with the functions across to the test harness.  But with a helper method or two I'm sure I can get it set again.  For now a couple are back to using manually specified identifiers.

I'd definitely review all at once rather than by commit because I had to go back and forth on some things.  The biggest changes (LabelSlot) are collapsed by default...


PwtKCL (commented):
Apart few comments, everything looks fine. The new tests looks sort of easier to understand than Cypress, it's amazing you put them out so fast.

neilccbrown:
Will address the items in the next PR, will merge for now.

PR #423, opened 2025-04-11
Fix selection-deletion under the new input events scheme

neilccbrown:
This adds some more tests for selection in slots using the new arrangement, and fixes those tests.

It's mainly about Firefox, which can do what I've called a "bad delete" which puts content into the parent div.  So we have to make sure we scan that and deal with it accordingly.  Let me know if the comments are clear enough; happy to write more if needed.

The tests all now pass and I think once this is merged (still into the input-events branch for now) I can bring the input-events branch up to date with main, and we can do a wider test on our test server.

### DIFF
--- a/src/components/LabelSlotsStructure.vue
+++ b/src/components/LabelSlotsStructure.vue
@@ -1,6 +1,7 @@
 <template>
     <div 
         :id="labelSlotsStructDivId"
+        :key="refactorCount"
         contenteditable="true"
         @keydown.left="onLRKeyDown($event)"
         @keydown.right="onLRKeyDown($event)"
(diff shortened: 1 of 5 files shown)


src/components/LabelSlotsStructure.vue
```diff
@@ -137,14 +138,53 @@ export default Vue.extend({
...
+                        sel = sel.replace(new RegExp(`[${UIDoubleQuotesCharacters[0]}${UIDoubleQuotesCharacters[1]}]`, "g"), STRING_DOUBLEQUOTE_PLACERHOLDER);
+                        sel = sel.replace(new RegExp(`[${UISingleQuotesCharacters[0]}${UISingleQuotesCharacters[1]}]`, "g"), STRING_SINGLEQUOTE_PLACERHOLDER);
+                        appendSel.append(sel + closing);
+                        closestDiv?.append(appendSel);
+                    }
+                }
+
+                const stateBeforeChanges = cloneDeep(this.appStore.$state);
+                // Must increment refactorCount in case the changed content doesn't trigger a noticeable refactor;
+                // we definitely need to completely redo all the slots if Firefox has deleted a bunch of nodes
```
PwtKCL, line 184: I'd just specify here the refactor is about the Vue component after "doesn't trigger a noticeable refactor" so it's clear what it is all about. Otherwise, you can add a small comment right at `refactorCount` declaration in the data.

src/helpers/editor.ts
```diff
@@ -288,7 +289,28 @@ export function getFrameLabelSlotLiteralCodeAndFocus(frameLabelStruct: HTMLEleme
     let foundFocusSpan = false;
     let ignoreSpan = !!delimiters;
     let hasStringSlots = false;
-    frameLabelStruct.querySelectorAll("." + scssVars.labelSlotInputClassName).forEach((spanElement) => {
+    // The container and intermediate divs can have relevant text if Firefox has done a "bad delete":
+    frameLabelStruct.querySelectorAll("." + scssVars.labelSlotInputClassName + ", ." + scssVars.labelSlotContainerClassName).forEach((spanElement) => {
```
PwtKCL, line 293: In the comment above, just for context, you may point at your explaining comment of LableSlot.vue

PwtKCL (commented):
It think it's good ! The explanation about "bad delete" is clear. I've only suggested small additions in the existing comments then all good to merge.

PR #433, opened 2025-04-24
State handling at start

PwtKCL:
In order to make the appearance of a saved state loading (from the browser storage) better when opening Strype, the default state with pre-existing frames is _only_ loaded if there is no existing saved state in the browser's storage.
(Note that there is _still_ a initial state being loaded, but it doesn't contain any other frames but the root and sections.)

### DIFF
--- a/src/store/store.ts
+++ b/src/store/store.ts
@@ -1316,7 +1316,6 @@ export const useStore = defineStore("app", {
         },
 
         saveStateChanges(previousState: (typeof this.$state)) {
-            console.log("in savestatchanged");
             this.isEditorContentModified = true;
             // Saves the state changes in diffPreviousState.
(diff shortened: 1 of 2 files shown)


src/store/store.ts
```diff
@@ -1291,6 +1316,7 @@ export const useStore = defineStore("app", {
         },
 
         saveStateChanges(previousState: (typeof this.$state)) {
+            console.log("in savestatchanged");
```
neilccbrown, line 1319: Haha, not just me!

neilccbrown (commented):
Looks fine, just one line to remove.

src/store/store.ts
PwtKCL: oops :)

PR #435, opened 2025-04-28
Auto-close error popups

### LINKED ISSUE #389: Add a close button on error popups
Sometimes an error popup can get in the way to the point where you can't actually see behind it (or get to the point of the error to fix it):

Image

I think the best solution is to add a close button which prevents that pop-up automatically re-showing until the next run of the code (but we keep the error marker in the frame to indicate the location).

PwtKCL:
This PR is an answer to fix #389 , which requested a mechanism for closing popups that were "in the way" sometimes depending on the website content and viewport.

Having a close button was a bit complex to realise for frame error popups because it messed up the styling (for frames, we have one popup object which style is dynamically modified according to the type of error).

So an alternative, a timeout to auto-close the popup after 3 seconds has been implemented.
(It concerns the popups for frames and label slots errors).

### DIFF
--- a/src/components/LabelSlot.vue
+++ b/src/components/LabelSlot.vue
@@ -43,6 +43,8 @@
             :content="errorMessage"
             custom-class="error-popover modified-title-popover"
             placement="bottom"
+            @shown="setAutoCloseErroPopup"
+            @hide="clearAutoCloseErrorPopupHandle"
         >
         </b-popover>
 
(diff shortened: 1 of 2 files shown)


src/components/LabelSlot.vue
```diff
@@ -43,6 +43,8 @@
             :content="errorMessage"
             custom-class="error-popover modified-title-popover"
             placement="bottom"
+            @shown="setAutoCloseErroPopup"
```
neilccbrown, line 46: Same again in this file

src/components/LabelSlot.vue
```diff
@@ -335,6 +340,25 @@ export default Vue.extend({
             return this.isFirstChange && this.appStore.isErroneousSlot(this.coreSlotInfo);
         },
 
+        clearAutoCloseErrorPopupHandle(event: BvEvent) {
+            const closeErrorPupHandleId = document.getElementById(event.componentId??"undef" )?.getAttribute("popup-close-tiemouthandle");
```
neilccbrown, line 344: And here.

neilccbrown (commented):
Looks fine overall, but I think it's worth fixing the typos.

src/components/LabelSlot.vue
PwtKCL: When inattention meets copy/paste... Sorry about that, and thanks for pointing them out!
I will push the corrections and merge afterwards.

PR #436, opened 2025-05-01
Fix non-ASCII character input in slots

### LINKED ISSUE #220: Foreign language input issues
In GitLab by @PwtKCL on Jun 18, 2024, 14:46

Observed on Windows, with a Chinese IME (over French layout): typing the double quotes won't be seen as our string wrapping for string literals.

### LINKED ISSUE #287: Cannot enter # in string on Mac
In GitLab by @mkolling on Sep 16, 2024, 20:29

When I try to enter a # character in a string on a mac (Option-3), nothing happens.
On a German keyboard (on mac too), "[" couldn't be added either.

neilccbrown:
I updated the input-events branch (which you have already seen all of and approved) by merging the main branch.  The only conflicts were related to import lines (where main had added some things and input-events had removed some other things) so if all the tests pass then I think it should be fine to merge without really needing any reviewing (since you've seen it all before).

For posterity: this merges the input-events branch which changes our slot input from acting on the keydown event to acting on the input event.  This lets us accept non-ASCII input including things like dead key input, alt-key shortcuts on German keyboards and things like Pinyin in Chinese.  This involved a reasonably large rewrite, especially around things like selection, and slot refactoring.  We now use a library called Playwright for some tests (which lets us test across browser, and in particular lets us test shift-selection which Cypress cannot).

Fixes #220 
Fixes #287


PwtKCL:
As tets passed, I merge it.

PR #437, opened 2025-05-02
Replaced timeouts by listening "escape" key hit to close error popups

PwtKCL:
After discussing in meeting the timeout to close the error popups, a better approach has been suggested: to listen to "escape" key instead. (see #435)

This PR also bumps the version for a following new release.

### DIFF
--- a/src/components/LabelSlot.vue
+++ b/src/components/LabelSlot.vue
@@ -649,7 +649,7 @@ export default Vue.extend({
         
         onEscKeyUp(event: KeyboardEvent) {
             // When an error popup is showing, the popup takes precedence
-            if([...document.getElementsByClassName("popover b-popover error-popover")].map((popup) => (popup as HTMLDivElement).style.display != "none").some((popupShowing) => popupShowing)) {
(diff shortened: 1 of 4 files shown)


src/components/LabelSlot.vue
```diff
@@ -672,6 +648,11 @@ export default Vue.extend({
         },
         
         onEscKeyUp(event: KeyboardEvent) {
+            // When an error popup is showing, the popup takes precedence
+            if([...document.getElementsByClassName("popover b-popover error-popover")].map((popup) => (popup as HTMLDivElement).style.display != "none").some((popupShowing) => popupShowing)) {
```
neilccbrown, line 652: I think you could change .map to .some, and then remove the trailing .some call altogether?

neilccbrown (commented):
Looks fine just one suggestion but I'll leave it up to you.

src/components/LabelSlot.vue
PwtKCL: Yes, really good point... !

PR #448, opened 2025-05-20
Moveable brackets

### LINKED ISSUE #284: Make deleting a bracket shunt it later in the expression?
I had code like this:

image

I wanted to delete the bracket (as you would in text) to then place it later in the expression.  But of course Strype doesn't permit this.  It made me wonder: what if pressing delete before a bracket did the following: move all content at the same or greater bracketing level inside the bracket.  Using | to represent the cursor, here's some examples of what pressing delete would do:

    foo(|)bar    --->    foo(bar)
    foo(1,|)bar(2) --->  foo(1, bar(2))
    foo(bar(|)3) + 4 ---> foo(bar(3)) + 4
    foo([1,|]2).bar() ---> foo([1,2]).bar()

It seems like a small item, but I think it would help with a lot of the awkwardness of working with brackets in Strype.

PwtKCL:
This branch contains the implementation of a new feature to allow "pushing" one bracket of a pair in the Strype editor.
It fixes #284  (see details of the issue for explanations).


src/components/LabelSlotsStructure.vue
```diff
@@ -616,6 +628,130 @@ export default Vue.extend({
...
+                    pushBracket = (neighbourSlot as BaseSlot).code.length > 0 || !new RegExp("^" + neighbourContainerSlotIndexToCheck + "$|," + neighbourContainerSlotIndexToCheck + "$").test(neighbourSlotInfosToCheck.slotId);
+                }
+            }
+            
+            if(pushBracket){
+                // To faciliate the code refactoring, we follow this approach: we clear off the full label slot structure, put the amended content in a new single slot,
+                // and refactor the slots to let our refactoring splitting up operators, brackets etc as normal.
+                let resultingCode = "", newCursorPosition = 0;
+                const neighbourToPushInSlotSpanId = getLabelSlotUID({...slotInfos, slotId: neighbourSlotInfosToCheck.slotId});
+                document.getElementById(this.labelSlotsStructDivId)?.querySelectorAll("." + scssVars.labelSlotInputClassName + ",." + scssVars.labelSlotMediaClassName).forEach((span) => {
```
neilccbrown, line 710: The media slot mention reminds me: does this do the right thing with media literal slots?

src/components/LabelSlotsStructure.vue
```diff
@@ -616,6 +628,130 @@ export default Vue.extend({
...
+                                const quote = (UIDoubleQuotesCharacters.includes(spanText)) ? "\"" : "'";
+                                spanText = quote;
+                            }
+                            resultingCode += spanText.replace("\u200b","");
+                        }
+                    }
+                
+                });
+                const stateBeforeChanges = cloneDeep(this.appStore.$state);
+                this.appStore.frameObjects[slotInfos.frameId].labelSlotsDict[slotInfos.labelSlotsIndex].slotStructures.fields.splice(1);
```
neilccbrown, line 746: Maybe add a comment here to explain the various indexes involved?

neilccbrown (commented):
Just a couple of comments to look at, otherwise I think should be fine once tests pass.

PR #453, opened 2025-05-21
Fixed a bug where copying strings would copy as Unicode quotes rather than ASCII quotes

### LINKED ISSUE #451: Issue with copying string literals from Strype
I am not sure if we always had the issue: if a string literal is copied from Strype, the quotes aren't copied as straight quotes but as decorative quotes, as we use in Strype UI.
For example, I can copy some code from the Strype slots and the text generated will be:
“aa” instead of "aa"

neilccbrown:
This also revealed an issue where pressing backspace while you have backwards-selected an empty string in a slot would delete the frame.

Tests are now present for both cases.

Fixes #451


PwtKCL (commented):
Just a console.log that can be removed, then all good to merge.

Issue #449, opened 2025-05-21
Graphics: pasting an image doesn't convert it (Windows/Chrome)


PwtKCL:
Using Chrome on a Windows machine (my KCL laptop), pasting an image inside a code expecting an image (like `set_background()`) does not fully work.
The image is detected by Strype: the wrapper `load_image()` is inserted fine, but the image itself isn't showing as an image, the base64 string shows instead.

![Image](https://github.com/user-attachments/assets/18cb9b62-e61b-4b20-be9f-0f095b006b07)

(The same works fine in FF however.)

PwtKCL:
Another note, when I copied this text in an empty slot in Chrome, the image was generated totally fine.
`set_background(load_image("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAgAAAAIAQMAAAD+wSzIAAAABlBMVEX///+/v7+jQ3Y5AAAADklEQVQI12P4AIX8EAgALgAD/aNpbtEAAAAASUVORK5CYII"))`

Issue #450, opened 2025-05-21
Graphics: copy slots containing media omits the media


PwtKCL:
Might a by design behaviour, but it could probably be overcome: testing on FF/Windows, if I copy slots from a frame that contains media, the media are not copied at all.

Example in Strype (the highlighted slots are being copied) :

![Image](https://github.com/user-attachments/assets/dd4eedc7-aaaa-43e6-9ee4-de26143029a9)

results in having this code pasted :

![Image](https://github.com/user-attachments/assets/81416be9-96a2-4223-879b-78a657d3aee8)

PR #455, opened 2025-05-21
Fix bugs with media literals

neilccbrown:
Fixes the issue with not copying media literals, and separately with not being able to paste them on Chrome.  Well spotted!  Also adds a test for the former, but I couldn't reproduce the permissions issue in the Playwright tests, only on desktop which is a bit annoying.


PwtKCL:
Fixes #449 and #450

PwtKCL (approved):

PR #456, opened 2025-05-22
New SPY format for saving

neilccbrown:
As previously discussed, this PR changes our SPY saving to save a valid Python file, with special comments for metadata (e.g. graphics divider positions) and special markers for invalid/unparseable code.

The PR looks huge, and it is big but it's not quite as bad as it looks from the numbers because a lot of it is the test framework: there's a bunch of manually specified tests (which you can skim, but might be useful to see what the format looks like; they're near the bottom), but also a randomised test framework that makes up a fairly garbage bit of code and checks that it can be entered, saved and loaded again.  I would definitely look through just the final version, as some of the commits along the way went back and forth on things.

The two main changes in the non-test code are:
 - The parser.ts file now looks for invalid Python while saving an SPY, and transforms it into valid Python using various ___strype-prefixed escapes.  The pythonToFrames.ts file has to spot these and restore the code accordingly, although that's fairly straightforward.
 - I've changed how the pythonToFrames.ts file deals with comments and blanks.  Skulpt drops these while parsing, so previously we had to keep track of them, and re-insert them at the appropriate points.  I started doing the same for the special comments indicating disabled frames and it became a bit of a nightmare.  So I've changed approach: we turn all blank lines and comments into actual non-comment Python and transform them when we see them in the parse tree.  Disabled frames become non-disabled but I just record the line numbers of disabled frames so that we can disable them once parsed.  This made the whole thing a lot easier and the code a lot cleaner.

We can still load the old .spy files because our logic was already: if it begins with "{" treat it as JSON, otherwise treat it as Python.  We distinguish plain Python from SPY by looking for the special comment directives; if it has some, it's SPY, otherwise it's plain Python.

I guess one question is if there's any other metadata we need to save.  At the moment it's essentially just disabled frames, and the graphics divider positions.  I looked through the store and couldn't see anything else that is actually necessary to store when loading and saving; syntax errors can be regenerated, runtime errors don't need to be saved across sessions, all editor state (e.g. where the frame cursor is, the undo/redo) doesn't need to be saved to an external file.  But I might have missed something that we actually want.  (Project name?  I think that is effectively stored in the file name?)

We can discuss later if you like (and perhaps in the meeting).


src/main.ts
```diff
@@ -9,14 +9,21 @@ import vBlur from "v-blur";
...
 Vue.config.productionTip = false;
 
 // Version of the application to check code's import compatibility in the editor
 // note: that is not an offical software version of Strype, just a way to help us dealing with compatibility issues.
 // it MUST be kept as an integer matching value
 export const AppVersion = "5";
+// The version used in the new .spy file format.  We may increment this in future
+// if we introduce a breaking change to that file format.
+export const AppSPYSaveVersion = "1";
+export const AppName = "Strype";
```
PwtKCL, line 23: Not sure why we need that?

tests/cypress/e2e/load-save.cy.ts
```diff
@@ -0,0 +1,242 @@
...
+        });
+
+        checkDownloadedFileEquals(spy.replaceAll("\r\n", "\n"), "My project.spy", true);
+    });
+} 
+
+describe("Loads and re-saves fixture files", () => {
+    if (Cypress.env("mode") === "microbit") {
+        // TODO instead issue a warning dialog when loading from another platform:
+        return;
```
PwtKCL, line 131: I'm not entirely clear why the microbit version can't be tested yet? With the right file, the current state of the SPY files is compatible with microbit right?

PwtKCL (commented):
A few comments about some parts of the code, but otherwise seems fine.
I suppose something need to be added in the PR anyway because the tests currently failed, so let's wait for the other push iteration.

src/App.vue
```diff
@@ -1186,10 +1195,20 @@ export default Vue.extend({
...
-                    (this.$refs[this.menuUID] as InstanceType<typeof Menu>).onFileLoaded(fileName, lastSaveDate, fileLocation);
-                    resolve();
+                    
+                    this.appStore.setDividerStates(
+                        loadDivider(s.headers["editorCommandsSplitterPane2Size"]),
+                        s.headers["peaLayoutMode"] !== undefined ? StrypePEALayoutMode[s.headers["peaLayoutMode"] as keyof typeof StrypePEALayoutMode] : undefined,
+                        loadDivider(s.headers["peaCommandsSplitterPane2Size"]),
+                        loadDivider(s.headers["peaSplitViewSplitterPane1Size"]),
+                        loadDivider(s.headers["peaExpandedSplitterPane2Size"]),
+                        () => {
```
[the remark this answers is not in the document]
neilccbrown: That's because the divider setting is asynchronous, and we want to do the notification after.  The intention was to mirror this existing code in Menu.vue (line 955):

```
this.appStore.setStateFromJSONStr(
                                        {
                                            stateJSONStr: reader.result as string,
                                        }
                                    ).then(() => fileHandles[0].getFile().then((file)=> this.onFileLoaded(fileHandles[0].name, file.lastModified, fileHandles[0])), () => {});
```

src/main.ts
neilccbrown: Probably not!  I thought it just might be better to have it in a constant so that it's synchronised between the load and the save code.

src/App.vue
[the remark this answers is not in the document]
PwtKCL: Sorry, I didn't read properly and thought the lines were statements rather than arguments... makes more sense

PR #459, opened 2025-05-23
Fix turtle/graphics interoperability

neilccbrown:
In the end, this was only really a few lines to fix, but this also adds some tests for keyboard and mouse input in turtle and Strype graphics, using Playwright.


PwtKCL (approved):
Great tests, all good to merge.

PR #478, opened 2025-06-05
Handle comments indentation from Python files + fix bug with wrapping with joint frames

PwtKCL:
For the first fix:
In Python, comments are not subject to indentation check, for example the code below is valid:
```
while True :
# a comment <-- should rather be aligned to next line, but Python accepts it
   doSomething()
```
However, this generates issues when we try to copy or open some Python code inside Strype, as comments are transformed for Skulpt to manage the code parsing. Therefore, Skulpt cannot know the transform comment was a comment.

We try to fix the indentation so that we can still open/paste the code.

For the second fix: see #482

### DIFF
--- a/src/helpers/pythonToFrames.ts
+++ b/src/helpers/pythonToFrames.ts
@@ -187,6 +187,45 @@ function transformCommentsAndBlanks(codeLines: string[]) : {disabledLines : numb
     const disabledLines : number[] = [];
     const transformedLines : string[] = [];
     const strypeDirectives: Map<string, string> = new Map<string, string>();
+
+    // A reference to the lines containing a comment block (that is, consecutive comment lines), see inline-method below for details.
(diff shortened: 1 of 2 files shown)


neilccbrown:
It looks like there's one test which has failed which had a zero-indent comment at the end of a file which has been indented by the new logic.

I think there's two options: adjust the test to match our behaviour, or change the logic so that comments are only adjusted if they are at an invalid position, and treat 0 as a valid position at the end of the file.  I slightly favour the latter but I don't have a strong preference, as I think comments at the end of blocks or files are reasonably rare.

PwtKCL:
The following amendment to the code is to implement the second option you suggested (for comments at the end of the code, change to the same indent than before unless it's at 0 - and when we detect 0 all other following comments need to be at 0 too).

PR #498, opened 2025-06-10
Fix Playwright tests

neilccbrown:
As mentioned, I realised if the Playwright tests were failing or timing out, Github wasn't noticing.  This fixes that, and also fixes a few (existing) test failures that it was showing up.  I would view all-at-once because I went round in circles for a bit trying to find the issue.


playwright.config.ts
```diff
@@ -18,10 +18,10 @@ export default defineConfig({
     outputDir: "./tests/playwright/test-results",
     /* Fail the build on CI if you accidentally left test.only in the source code. */
     forbidOnly: !!process.env.CI,
-    /* Retry once on CI, as some of the random tests are slightly flaky */
-    retries: process.env.CI ? 1 : 0,
+    /* Retry twice on CI, as some of the random tests are slightly flaky */
+    retries: process.env.CI ? 2 : 0,
     fullyParallel: true,
-    workers: process.env.CI ? 1 : 4,
+    workers: process.env.CI ? 1 : 1,
```
PwtKCL, line 24: Was the change temporary? If not, then would it just be 1 without conditional?

tests/playwright/e2e/graphics.spec.ts
```diff
@@ -11,7 +11,9 @@ test.beforeEach(async ({ page, browserName }, testInfo) => {
     if (browserName === "webkit" && process.platform === "win32") {
         // On Windows+Webkit it just can't seem to load the page for some reason:
         testInfo.skip(true, "Skipping on Windows + WebKit due to unknown problems");
-    }    
+    }
+
+    testInfo.setTimeout(90000); // 90 seconds
```
PwtKCL, line 16: Maybe just tell why a new timeout is needed here?

tests/playwright/e2e/load-save-random.spec.ts
```diff
@@ -28,13 +28,19 @@ function createBrowserProxy(page: Page, objectName: string) : any {
...
     // (pressing right out of a comment frame puts the cursor at the beginning and makes a frame cursor).
     // Since it works in the real browsers, and on Webkit and Firefox, we just skip the tests in Chromium
-    test.skip(testInfo.project.name == "chromium", "Cannot run in Chromium");    
+    test.skip(testInfo.project.name == "chromium", "Cannot run in Chromium");
+    if (browserName === "webkit" && process.platform === "win32") {
+        // On Windows+Webkit it just can't seem to load the page for some reason:
+        testInfo.skip(true, "Skipping on Windows + WebKit due to unknown problems");
+    }
+    
+    testInfo.setTimeout(90000); // 90 seconds
```
PwtKCL, line 43: Also here, maybe comment why extra time is required?

PwtKCL (commented):
Sorry for checking the PR late. It seems a lot of issues were indeed due to the promises mechanisms used by the tests. 
I have just a few comments otherwise good to merge.

playwright.config.ts
neilccbrown: The parallel thing is awkward; ideally I'd run them parallel but the tests for downloading and checking images download to the same directory and in some cases to the same file name so they can't run in parallel.  But the other side could...  I think you're right I should just set it to 1, but maybe I can investigate running some-but-not-all tests in parallel.

Issue #483, opened 2025-06-06
Project's layout isn't showing on reload


PwtKCL:
Change the sizes of the pane in the current project in Strype (for example, making the frame commands/PEA very narrow).
Then reload the page: the project loads, but the layout isn't maintained.

The problem is the same if I load a project from a given location.
(Note: it seems that some parameters can be restored, for example, the type of layout.)

PR #509, opened 2025-06-16
Fix cursor position and after click, and loading/saving of dividers

neilccbrown:
This fixes two bugs:
 - The cursor position not being right after clicking.  It put the cursor position after the invisible zero-width space in an empty slot before bracket, which messed things up.  Instead it should put it before that zero-width space.  (All the other code with keyboard clicking etc accounts for this, but not if you clicked the left-hand half of an operator or bracket.)
 - Dividers were not being saved properly.  This was down to a repeated name overwriting divider info, and an issue with Vue reactivity inside a splitter object.  From what GPT tells me, Vue reactivity works with string keys but we were using int keys that were auto converted to strings.  Certainly, when I swapped the enum to have string values rather than ints, the tests passed so that does seem to be the right fix.  We were already saving with string names so it doesn't affect .SPY loading.  Loading existing local browser state will momentarily forget the dividers in the first time after deploying but I think users will survive the one-off issue.

In the end, both bugs were short fixes.  But I added a bunch of tests to show up the errors and that's actually the bulk of the commits!  Might need a little more manual testing with the dividers to be sure we're happy, but I think it's not a critical bug even if needs further work in future.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1201,7 +1201,7 @@ export default Vue.extend({
         onStrypeCommandsSplitPaneResize(event: any, useSpecificPEALayout?: StrypePEALayoutMode){
             // Save the new size of the RHS pane of the editor/commands splitter
             if(this.appStore.editorCommandsSplitterPane2Size != undefined) {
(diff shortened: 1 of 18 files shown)


PwtKCL (approved):
For the dividers, the issue was #483

PR #510, opened 2025-06-17
Add parameter prompts for methods

### LINKED ISSUE #506: Add param prompts for object method calls
Previously with Skulpt if you had code like:

```
a = Actor("img.jpg")
a.move()
```

Then we couldn't provide a parameter name prompt for move because Skulpt didn't have the necessary reflection info.  But now, with TigerPython, I think we should be able to provide that info in many cases by inferring the type of "a" (i.e. the object on the LHS).

neilccbrown:
We have had support for function parameter prompts for a while, i.e. you type `set_background()` and it shows the parameter name in grey in the brackets.   But we didn't have the same for `fish.move()` because Skulpt had no support for dynamically telling us the param names of an object method.  However, now we're use TigerPython, it can work out the type of fish, and thus we can look at the list of methods and find move and its parameter names.

There are a few limitations of TigerPython that we can look to improve.  First, it doesn't tell us which parameters have default values, which means we (wrongly) show defaults as missing.  E.g. if you have `get_touching(tag=None)` then it shows `tag` even though it's not required.  Second, it doesn't seem to have inference for fully qualified items, e.g. if you `import strype.graphics` and type `strype.graphics.Actor` it doesn't seem to pick it up.  I plan to go and add support for that to Tiger Python afterwards, but I think this PR is probably still worth merging and releasing in the mean time; thinking of the workshop in particular, I don't think any of the methods we use have a default param that we won't be filling in.

Fixes #506


src/autocompletion/acManager.ts
```diff
@@ -504,14 +526,30 @@ export async function calculateParamPrompt(context: string, token: string, param
             return "\u200b";
         }
     }
+
+    if (context) {
+        // See if TigerPython can infer the type of the content before the .
+        const parser = new Parser();
+        const userCode = parser.getCodeWithoutErrors(frameId, true);
+        await tpyDefineLibraries(parser);
+        // So we get code out which is a partial expression and may not be valid at top-level.  Thus we wrap it in:
```
PwtKCL, line 535: Just for clarity here I would amend slightly the comment to point at the "_context_ code", because when I read it I got confused with "userCode" that is used before, and which is not what the comment talks about.

src/helpers/editor.ts
```diff
@@ -1460,7 +1463,39 @@ export const getSameLevelAncestorIndex = (slotId: string, sameLevelThanSlotParen
     return parseInt(slotId.split(",")[ancestorLevels -1]);
 };
 
-const FIELD_PLACERHOLDER = "$strype_field_placeholder$";
+const FIELD_PLACEHOLDER_START = "$strype_field_placeholder";
```
PwtKCL, line 1466: With the work you've done in mind, the new format makes a lot of sense. But with time we might wonder why we needed to keep the underlying code within the placeholder: maybe just add a small comment to tell why this is important to keep the underlying code in.

src/helpers/editor.ts
```diff
@@ -1460,7 +1463,39 @@ export const getSameLevelAncestorIndex = (slotId: string, sameLevelThanSlotParen
...
+    const newInput =
+        input.substring(0, startIndex) +
+        decoded +
+        input.substring(endIndex + FIELD_PLACEHOLDER_END.length);
+
+    // Recurse to handle more placeholders:
+    return transformFieldPlaceholders(newInput);
+}
+
 export const IMAGE_PLACERHOLDER = "$strype_image_placeholder$";
```
PwtKCL, line 1499: Thanks to my bad spelling again, this has passed through "IMAGE_PLACERHOLDER" (which should be "IMAGE_PLACEHOLDER"). Sorry !!

src/helpers/editor.ts
```diff
@@ -1460,7 +1463,39 @@ export const getSameLevelAncestorIndex = (slotId: string, sameLevelThanSlotParen
...
+        const end = s.indexOf(FIELD_PLACEHOLDER_END);
+        return s.slice(end + FIELD_PLACEHOLDER_END.length);
+    }
+    else {
+        return s;
+    }
+}
+export function transformFieldPlaceholders(input: string) : string {
+    input = input.replaceAll(STRING_SINGLEQUOTE_PLACERHOLDER, "'");
+    input = input.replaceAll(STRING_DOUBLEQUOTE_PLACERHOLDER, "\"");
```
PwtKCL, line 1479: Thanks to my bad spelling, these had passed through "STRING_SINGLEQUOTE_PLACERHOLDER" (which should be "STRING_SINGLEQUOTE_PLACEHOLDE"), and the same applies for the other one... Sorry !!

src/helpers/pythonToFrames.ts
```diff
@@ -586,7 +586,7 @@ function replaceMediaLiteralsAndInvalidOps(s : SlotsStructure) : SlotsStructure
     return s;
 }
 
-function fromUnicodeEscapes(input: string): string {
+export function fromUnicodeEscapes(input: string): string {
```
PwtKCL, line 589: Now that this function is exported, I wonder if it shouldn't be moved to /helpers/commons.ts as it has a broad purpose. Up to you, it's not very important.

tests/cypress/e2e/param-prompts-objects.cy.ts
```diff
@@ -0,0 +1,27 @@
...
+describe("Parameter prompts objects", () => {
+    // Each item is a triple: the module, the function name within the module, the list of param names
+    const rawFuncs : [string | null | [string, string], string, string[]][] = [];
+    rawFuncs.push([null, "str(8).center", ["width"]]);
+    if (Cypress.env("mode") !== "microbit") {
+        rawFuncs.push(["strype.graphics", "Actor(s).move", ["distance"]]);
+        rawFuncs.push(["strype.graphics", "Actor(t).set_location", ["x", "y"]]);
+        rawFuncs.push(["strype.graphics", "get_background().draw_circle", ["centre_x", "centre_y", "radius"]]);
+        // TODO this should be ["text", "x", "y"] but TigerPython doesn't (yet)
+        // give us info on which functions have default values, so it shows the full set:
```
PwtKCL, line 20: For future reference and ease of checking, we can specify the current version of TigerPython so we don't need to check dates in some later time.
(Although, I'm not sure if TigerPython keeps a clear versioning, but well...)

PwtKCL (commented):
I have only a few comments-related changes that you can address if need be.
Otherwise, all good to merge.

PR #528, opened 2025-06-24
A few various fixes (most from tracked issues).

PwtKCL:
This PR contains some fixes for logged issues (mainly). Most issues are related to selection/editing in Strype, and one is for Google Drive.
There might be other fixes to do later but that's a decent size for review now.

See each commit for details.

### DIFF
--- a/src/components/LabelSlot.vue
+++ b/src/components/LabelSlot.vue
@@ -637,8 +637,7 @@ export default Vue.extend({
             const isArrowUp = (event.key == "ArrowUp");
            
             // Check if we can reach another VISUAL line in a comment (this method returns false if we're not in a comment frame)
-            const canReachAnotherCommentLine = checkCanReachAnotherCommentLine((this.frameType == AllFrameTypesIdentifier.comment), isArrowUp, document.getElementById(getLabelSlotUID(this.coreSlotInfo)) as HTMLSpanElement); /*&& isCommentFrame && ((isArrowUp && slotContentToCursor.includes("\n")) || (!isArrowUp && slotContentAfterCursor.includes("\n")))*/
-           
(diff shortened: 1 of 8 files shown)


src/components/LabelSlot.vue
```diff
@@ -651,9 +650,28 @@ export default Vue.extend({
...
+                    // Because of Firefox usual problems, we can't reliably use the document selection to infer where we are within the span.
+                    // Instead, we let it be handled by the brower and check the selection again in a short time: if it's different, it means 
+                    // the up/down arrow was used by the browser to change the position of the cursor within the comment and we don't do anything else. 
+                    // However if the selection remains the same, then we can assume the cursor hasn't changed because there is no line to get to,
+                    // so we can consequently get out the frame.
+                    const prevSelectionRange = document.getSelection()?.getRangeAt(0).cloneRange();
+                    setTimeout(() => {
+                        try{
+                            const currentSelectionRange = document.getSelection()?.getRangeAt(0).cloneRange();
+                            if(prevSelectionRange && currentSelectionRange?.startOffset==prevSelectionRange.startOffset && currentSelectionRange?.endOffset==prevSelectionRange.endOffset){
```
neilccbrown, line 667: Do we also need to check that the node for the selection is the same?  E.g. if they are are at the start of one empty comment and click at the start of the empty next comment, will this code fire?

neilccbrown (commented):
Generally all good, just one thing to check.

src/components/LabelSlot.vue
PwtKCL: No, because this code (inside the setTimeout method) is only called upon up/down arrows key actions, and if we are in a comment.

PR #531, opened 2025-06-27
Fix up method parameter prompts

neilccbrown:
This includes all the changes seen in the meeting yesterday, plus all the fixes we then agreed on.  It's a bit long but I would review all-at-once, because the individual commits had a lot of back-and-forth.  You can probably just skim the test code and focus on the main code changes.


PwtKCL (commented):
Very impressive! Just got a few typos spotted (and 2 of the comments cancel each other, so it's 3 really) then good to merge

PR #533, opened 2025-06-27
A few more things on comments (multiline)

PwtKCL:
The main commit is the first one that handles the Python transformation for us to interpret the multilines comments.
As this part is quite a technical one, I hope I haven't messed up existing code, and considered most cases.
I've tested reloading a SPY file, and copying some Python that is decently formatted to test.

Other commits are just small fixes.


src/helpers/pythonToFrames.ts
```diff
@@ -305,10 +306,43 @@ function transformCommentsAndBlanks(codeLines: string[], format: "py" | "spy") :
             }
             checkRearrangeCommentsIdent();
         }
+        else if(codeLines[i].trim().startsWith("'''") || codeLines[i].trim().endsWith("'''")){
+            // Note: we ignore the triple quotes *strings* in our code.
+            const indexOfFirstToken = codeLines[i].indexOf("'''");
+            if(codeLines[i].trim().startsWith("'''") && codeLines[i].trim().endsWith("'''") && codeLines[i].trimStart().length >= 6){
+                // If we have triple quote token both at the start and at the end, it's actually a single line comment...
```
neilccbrown, line 313: It can also be the end of one and start of next...  (I don't mind too much if we don't support this.)

neilccbrown, line 313: It can also be the end of one and start of next...  (I don't mind too much if we don't support this.)

src/helpers/pythonToFrames.ts
```diff
@@ -305,10 +306,43 @@ function transformCommentsAndBlanks(codeLines: string[], format: "py" | "spy") :
...
+            }
+            else{
+                // A multlines comment: we treat every line as part of a multilines comment,
+                // further parsing will notice we are in inside a multilines comment.
+                // We need to set a flag that we are starting/ending a multilines comment for this parsing context.
+                // The comment token (''') needs to be kept for later knowing when we start/end a comment.
+                // (note: we need a specific comment SPY token to make a distinction between comments and multine comments, 
+                // otherwise a code like "#'''" would become ambiguous once parsed...)
+                const transformedCodeContent = (isParsingMultiLinesComment) 
+                    ? mostRecentIndent + STRYPE_MULTILINES_COMMENT_PREFIX + toUnicodeEscapes((codeLines[i].slice(0, mostRecentIndent.length).trim().length == 0) ? codeLines[i].slice(mostRecentIndent.length) : codeLines[i].trimStart())
```
neilccbrown, line 326: I think slicing by mostRecentIndent is wrong if the multiline string isn't indented beyond most recent indent.  (This may not be invalid, consider

```
def foo():
     if blah:
          pass
     """ 
     hi
     """
```
The most recent indent when we encounter the string is 8, but the multiline literal is indented 4.

src/helpers/pythonToFrames.ts
```diff
@@ -305,10 +306,43 @@ function transformCommentsAndBlanks(codeLines: string[], format: "py" | "spy") :
...
+                isParsingMultiLinesComment = !isParsingMultiLinesComment;
+            }
+        }
         else {
-            transformedLines.push(codeLines[i].trimEnd());
-            mostRecentIndent = getIndent(codeLines[i].trimEnd());
-            checkRearrangeCommentsIdent();
+            // Any other valid code, except if we are in the context of multilines comment (see above)
+            if(isParsingMultiLinesComment){
+                transformedLines.push(mostRecentIndent+STRYPE_MULTILINES_COMMENT_PREFIX + toUnicodeEscapes((codeLines[i].slice(0, mostRecentIndent.length).trim().length == 0) ? codeLines[i].slice(mostRecentIndent.length) : codeLines[i].trimStart()));
```
neilccbrown, line 339: Same note here about slicing by mostRecentIndent; what if they do that thing where the string content is less than the indent of the opening quote?

src/helpers/pythonToFrames.ts
```diff
@@ -720,16 +754,17 @@ function makeAndAddFrameWithBody(p: ParsedConcreteTree, frameType: string, keywo
     return {s: s, frame: frame};
 }
 
-// Process the given node in the tree at the current point designed by CopyState 
+// Process the given node in the tree at the current point designed by CopyState,
+// and maintain an object, extraInfos, for things that need to be deferred (for example, multilines comments).
 // Returns a copy state, including the frame ID of the next insertion point for any following statements
-function copyFramesFromPython(p: ParsedConcreteTree, s : CopyState) : CopyState {
+function copyFramesFromPython(p: ParsedConcreteTree, s : CopyState, extraInfos: {multiLinesCommentContent: string | null} = {multiLinesCommentContent: null}) : CopyState {
```
neilccbrown, line 760: I would put multiLinesCommentContent into CopyState, it's already doing this kind of role (keeping track of state across the parsing).

neilccbrown (commented):
Generally it's a good approach I think.  Just a few things to take a look at.

src/helpers/pythonToFrames.ts
PwtKCL: As discussed, I'll look how to handle such case: `'''a first comment''' ''' another comment on same line'''`. For case where we detect something in between, for the moment we won't handle it with intelligence to generate strings.

src/helpers/pythonToFrames.ts
PwtKCL: Good point.

src/helpers/pythonToFrames.ts
PwtKCL: The code would only slice by mostRecentIndent if the trimmed slice's length is 0 (only spacings). If not, it will remove spacing and start at the first none-spacing character. It might not be exact in indentation as wanted by user, 
At the moment this code; 
image
will render
image

Should it render
image

But then what do we do with the third line?

src/helpers/pythonToFrames.ts
PwtKCL: Actually, the part that is run if the condition is true, in this ternary operator, is for the closing comment.
We parse by the ''' token when we start parsing a multilines comment.
However; trying your example I thought I wasn't handling the indentation properly because the comment would placed inside the if block.
The problem was not coming from indents, but from the cleanup (checkRearrangeCommentsIdent()) that would push those comments to the same alignment than in the if block.

src/helpers/pythonToFrames.ts
PwtKCL: See new version committed for details: the indentation inside the comments shouldn't matter, only the indentation of the starting ''' if it starts a line does.

src/helpers/pythonToFrames.ts
```diff
@@ -9,6 +9,8 @@ import FrameContainer from "@/components/FrameContainer.vue";
 
 const TOP_LEVEL_TEMP_ID = -999;
 
+const parsedTripleQuotesStrRegex = /^''.*''$/s;
```
neilccbrown, line 12: Maybe make a comment here about why there are only two quotes: I think it's because this is the string content once the outer quotes are processed by Skulpt?  Also, what about triple double-quote strings, i.e. """ blah """ ?

neilccbrown (commented):
I think broad gist looks fine, but I think triple double-quotes are not yet being handled, only triple single-quotes?  Unless I missed something.  But Python can have either as I understand it.

PwtKCL:
> I think broad gist looks fine, but I think triple double-quotes are not yet being handled, only triple single-quotes? Unless I missed something. But Python can have either as I understand it.

That's correct. I didn't implement the triple quotes (in honesty I didn't think of it, when I copied your comment I thought errors came from using them instead of single quotes). I will let it out for the time being, and can always work on it separately later.

Issue #535, opened 2025-06-28
Loading share projects shouldn't show it is saved somewhere (FS)


PwtKCL:
When opening a shared proejct, for example from our workshop page, it shows as it if was saved on the file system.
It shouldn't.

PR #537, opened 2025-06-29
Small patch for shared projects

PwtKCL:
Generic non Google Drive shared projects were showing as saved on the File System instead of none, once loaded.
This is a simple fix, and because of the CAS North East worskhop tomorrow, I'll deploy it on our server.
(Note that the reference for the build will be 1.1.1 to only bring up this change live - I will use this branch (the one to merge) as a patch and tag it for reference internally.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "strype-editor",
-    "version": "1.1.0",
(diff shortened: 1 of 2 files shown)


PwtKCL:
This fixes #535

PR #544, opened 2025-07-04
Simplify the autocomplete code

neilccbrown:
This makes a few changes:
 - Remove the debugAC flag; I think we can live without it
 - Combine acRequested and showAC, and remove the debounc.  They seem to be almost identical, you can see the via the diff that they are usually manipulated together.  I think the only way this is different is that the AC shows immediately rather than waiting for updateAC; but since I've removed the debounce on updateAC (less necessary because the method it calls is now async, and we are using TigerPython rather than firing up Skulpt) I think this won't make any practical difference any more.

See what you think; it's not strictly a necessary change but I need to poke the AC code a little while doing the work on wrapping and comments so it seemed a good time to simplify things first.

### DIFF
--- a/src/components/LabelSlot.vue
+++ b/src/components/LabelSlot.vue
@@ -270,11 +270,7 @@ export default Vue.extend({
         errorHeader(): string{
             return this.appStore.getErrorHeaderForSlot(this.coreSlotInfo);
         },
-
-        debugAC(): boolean{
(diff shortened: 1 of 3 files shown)


PwtKCL (approved):
Yes the changes are sensible. I've virtually never used the debugAC flag, and it would not make sense to keep "in production" anyway - so it's good you spotted that.

PR #548, opened 2025-07-07
File IO implementation for support with Cloud-based projects.

PwtKCL:
The support for external file's IO was not supported by Skulpt because of browses' limitations to the local file system.
This PR adds functionality in both Skulpt and Strype to handle the user's code with regards to Python file IO: we want to allow it for IO operation on **cloud-based** files (for the moment, then, only Google Drive...)
(The access to the files to write/read can be done with a relative/absolute path, but the root of the file system will always be the current Strype's project location, and ".." is forbidden.)

This PR has a lot of back and forth commits and incremental changes. So I think it is easier to see it as a whole.
**Note:** because I needed to edit the Skulpt JS file, I have added a new file in our repository that shows the formatted version of the Skulpt file before my changes. The idea is to keep that file on the repository only for comparison purposes (it might be helpful to see the changes in Skulpt using that file rather than in Git, since the overall formatting would have changed).
I'm not too clear as to what we should do next. For now, the new Skulpt file is formatted and contains _specific comments to indicate where are my changes_. 
I thought after the code review we would revert the Skulpt file to a minimised version.


neilccbrown (commented):
All seems to make sense, just one item to document.  Good job!

PR #557, opened 2025-07-27
Changed the handling of up/down keys

neilccbrown:
Previously, even if a frame wrapped, up and down just went to the prev/next frame cursor, which was frustrating to navigate wrapped frames because it was quite laborious (and confusing) to navigate them.

This PR changes up/down to be handled by us and to look for the graphically horizontally-aligned position in the next span down (see spans.ts new file) to navigate to.  We only jump to the frame cursor from the top/bottom line.

Also changed where we handle up/down to be in LabelSlotStructure rather than LabelSlot since it needs full visibility of all available slots.

I also tweaked some things in the tests slightly because they were annoying me by taking so long to run.


src/components/LabelSlotsStructure.vue
```diff
@@ -668,6 +673,62 @@ export default Vue.extend({
...
+            if (!this.isFocused || !this.appStore.isEditing) {
+                return;
+            }
+            event.stopPropagation();
+            event.stopImmediatePropagation();
+            event.preventDefault();
+            
+            if (!(event.shiftKey || event.metaKey || event.altKey || event.ctrlKey)) {
+                const subSlots = this.$refs.labelSlots as InstanceType<typeof LabelSlot>[];
+                for (const subSlot of subSlots) {
```
PwtKCL, line 687: While is working, I wonder if it would not be more logical to use emit events rather than scanning all subslots to find which one did something. 
That is: if the subslot emits an event (with .$emit()) in handleUpDown(), we can add a v-on <that event type> in <LabelSlot> in here, and do the same but without having to loop.
It would mean, of course, that the logic of checking the other slots of that stuctures is down to the LabelSlot - which is probably not ideal.
So, I guess the 2 approaches can work, but I'm not sure the suggestion I made here is better just because it avoids scanning the array.

src/helpers/spans.ts
```diff
@@ -0,0 +1,120 @@
+function getCaretScreenPosition() : { x: number; y: number } | null {
```
PwtKCL, line 1: Just for information so we remember in future, can you add a very succint class-level explanation at the top of this file to explain its role.

src/components/LabelSlotsStructure.vue
neilccbrown: Yes it is a bit awkward.  I think the thing is that I need to know if the subslots captured the event or not.  It could maybe be solved by going back to handling up/down in labelslot and emitting the event to labelslotstructure.  I have a feeling there was an issue with this but I wrote this part of the code about a month ago before all the travelling.  So I think I may just live with it for now as it seems to work, but we can revisit this part in future if we need to edit it again.

Issue #578, opened 2025-09-30
Sharing project: we are always sharing a project even when not sharing


PwtKCL:
The current mechanism to handle the sharing functionality tries to share the project publicly before the user validates a choice (either sharing publicly or internal to the cloud drive).
The issue with that is that if the user cancels the action of sharing in Strype, the file will retain its shared status in the cloud unknowingly to the user.

So we should probably be more cautious by:
- checking if the file is already share,
- retain that status as a "previous state" in case we cancel
- if cancelled make sure the previous status of the file is maintained on the cloud drive...

Issue #551, opened 2025-07-16
Examples popup not showing if saving the current project


PwtKCL:
Modify a project in Strype (a project that is either only saved in the browser or coming from the cloud), and try to open an example.
The popup to ask for saving shows, but after saving, we get the opening dialog instead of the example dialog.

PR #580, opened 2025-10-02
First WIP for including OneDrive as one Cloud Drive to use with Strype

PwtKCL:
This is not an urgent PR. I wanted to start making some PR now with the changes before it is even worse in size.
The content of the PR here is already pretty big - sorry !

It contains changes to make a common CloudDriveHandler component that allows us to get "wrapping" component and logic holder between the different drives (at the moment, Google Drive and OneDrive). 
The dialog to load/save on the Cloud is slightly modified to allow space for a new Cloud and be nicer for other locales.

Also, the PR includes a new mechanism to remove public sharing when Strype sets it "in preparation" of a sharing. In previous versions of Strype, cancelling the sharing or sharing internally to the Cloud Drive let that public share active (#578).
There is another fix related to opening demos dialog not showing when saving a project beforehand (#551).

Finally, this PR includes the WIP for OneDrive. The integration of OneDrive is currently related to _personal accounts only_, I plan to work on work/school accounts later.
What is done:
- file opening (from picker, in any folder)
- file saving in /Strype and create of /Strype if not existent 
- folder selection for saving (partly: "save as" isn't working perfectly yet)
- sharing a file (needs more testing for internal sharing)
- opening a shared file (**see details below**)

Regarding sharing with OneDrive, I've looked up and it seems we cannot get a link that allows us to retrieve _the raw file content_ in a proper fashion. It is possible to get a link, but the link is _ephemeral_ and therefore mostly not suited for using in Strype.
So the other approach is to use the link that OneDrive creates upon setting a file shared, and allow users to download their file locally first.
That means, _for OneDrive_, we now show this when requesting a sharing link (the "OneDrive" showing there is dynamic based on the Cloud name, if we were to have the same scenario in the future with other Clouds):
image

Also, when users try to open the public share we show this error message:
image

After 3 seconds, we open the link in a new tab. For example, we then get this in the new tab;
image

### DIFF
--- a/src/components/CloudDriveHandler.vue
+++ b/src/components/CloudDriveHandler.vue
@@ -37,7 +37,7 @@ import OneDriveComponent from "@/components/OneDriveComponent.vue";
 import { generateSPYFileContent } from "@/helpers/load-save";
 import { AppSPYFullPrefix } from "@/main";
 
-// This enum is used for flaging the action taken when a request to save a file on Google Drive
+// This enum is used for flaging the action taken when a request to save a file on a Cloud Drive
 // has been done, and a file of the same name already exists on the Drive
 enum Actions{
     overwrite,
@@ -71,14 +71,6 @@ export default Vue.extend({
             Actions, // this is required to be accessible in the template
             saveExistingCloudProjectInfos: {} as SaveExistingCloudProjectInfos,
         };
-    },
-
-    created() {
-        // We do not add the specific Cloud Drive's components via the <template> part, to allow us typing them properly
-        // (see the cloud types file, it's a struggle to work that out).
-        // Anyway, there is no styling require for these components so it's not a loss...
-        // Google Drive
-
     },
 
     computed: {
@@ -176,7 +168,8 @@ export default Vue.extend({
             }            
         },
 
-        // Test the connection is still valid (and allow callbacks for success or failure of the test)
+        // Request the specific cloud drive component to test if the connection is still valid.
+        // We provide the compoent the callbacks methods for success or failure of the test which it will call accordingly.
         testCloudConnection(cloudTarget: StrypeSyncTarget){
             const cloudDriveComponent = this.getSpecificCloudDriveComponent(cloudTarget);
             cloudDriveComponent?.testCloudConnection(() => {
@@ -304,14 +297,17 @@ export default Vue.extend({
                         this.appStore.strypeProjectLocationAlias = folderName;
                     })
                     .catch((responseStatusCode) => {
+                        // The following error status codes were relevant for Google Drive. 
+                        // We keep them for the general cases, but add a catch up case for other error codes
+                        // that may be sent by other Cloud Drives.
                         // Connection issue?
                         if(responseStatusCode == 401 || responseStatusCode == 403){
                             this.proceedFailedConnectionCheckOnSave(cloudTarget);
                             return;
                         }
                         
-                        // Folder not found
-                        if(responseStatusCode == 404){
+                        // Folder not found and any other error (400+).
+                        if(responseStatusCode - 400 >= 0){
                             this.appStore.strypeProjectLocation = undefined;
                             this.appStore.strypeProjectLocationAlias = "";
                             this.appStore.projectLastSaveDate = -1;
@@ -416,7 +412,7 @@ export default Vue.extend({
             // Do something in case of connection failure depending on the reason for saving
             // normal saving: --> try to reconnect, if failed, then we stop synchronising to Google Drive
             // save to load + unload --> try to reconnect, if failed, stop sync + modal message
-            // Even if the user may signing again, we first make sure everything shows as "not syncing" in case the signing process is not completed
+            // Even if the user may sign-in again, we first make sure everything shows as "not syncing" in case the signing process is not completed
             // (because if the user just drop the signing action, we have no way to get events on that...)
             this.getSpecificCloudDriveComponent(cloudTarget)?.resetOAuthToken(); 
             this.updateSignInStatus(cloudTarget, false);
--- a/src/components/OneDriveComponent.vue
+++ b/src/components/OneDriveComponent.vue
@@ -85,6 +85,10 @@ export default Vue.extend({
             return "ee29b56f-8714-472f-a1c8-37e8551e3ec5";
         },
 
+        consumerTenantIdForPersonalAccounts():string {
+            return "9188040d-6c67-4c5b-b112-36a304b66dad";
+        },
+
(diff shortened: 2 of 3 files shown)


src/components/OneDriveComponent.vue
```diff
@@ -0,0 +1,840 @@
...
+        resetOAuthToken() {
+            this.oauthToken = null;
+        },
+
+        isOAuthTokenNotSet(){
+            return this.oauthToken == null;
+        },
+
+        testCloudConnection(onSuccessCallback: () => void, onFailureCallBack: () => void){
+            //TODO
```
neilccbrown, line 161: Is this still TODO?  It looks like it might be implemented.

src/components/OneDriveComponent.vue
```diff
@@ -0,0 +1,840 @@
...
+                    {
+                        method: "POST",
+                        headers: {"Authorization": `Bearer ${token}`,"Content-Type": "application/json"},
+                        body: "{name: 'Strype', 'folder':{}}",
+                    });
+                if (resp.ok) {
+                    const data = await resp.json() as BaseItem;
+                    return checkFolderDoneCallBack(data.id??"");
+                }
+            }
```
neilccbrown, line 531: Should we also handle other error codes?

src/components/OneDriveComponent.vue
```diff
@@ -0,0 +1,840 @@
...
+                    onSuccessCallback();
+                },(_) => {
+                    // We shouldn't have an issue at this stage, but if it happens, we just attempt to connect again
+                    onFailureCallBack();
+                });
+        },
+
+        checkIsFileLocked(existingFileId: string, onSuccess: VoidFunction, onFailure: VoidFunction) {
+            //TODO: is there an equivalent in OneDrive? If not, just return as never locked...
+            // // Following the addition of a locking file settings in Drive (Sept 2023) we need to check if a file is locked when we want to save.
```
neilccbrown, line 562: Does this comment refer to Google?  I guess you will revisit this method anyway, though.

src/localisation/de/de_main.json
```diff
@@ -63,12 +64,12 @@
...
     "errorUserDefinedVarMsg": "Benutzerdefinierte Variable\nEine Zuweisung zu dieser Variablen enthält einen Fehler",
     "EOFError": "Dieser Frame kann nicht leer sein.",
-    "GDriveSaveFailed": "Strype konnte nicht auf Google Drive speichern.",
+    "driveSaveFailed": "Strype konnte nicht auf {drivename} speichern.",
     "fileNameError": "Projekt- / Dateiname kann nur Buchstaben, Zahlen, Leerzeichen, Bindestrich oder Klammern beinhalten.",
     "gdriveConnectionSaveToLoadProjFailed": "Fehler beim Versuch, dass offene Projekt auf Google Drive zu speichern. Bitte wieder anmelden, um ein anderes Projekt zu laden.",
     "gdriveConnectionSaveToUnloadPageFailed": "Fehler beim Versuch, dass offene Projekt auf Google Drive zu speichern. Bitte wieder anmelden, um das Projekt zu speichern bevor Strype geschlossen wird.",
     "gdrivePermissionsNotMet": "Strype braucht midestens ein Zugriffsrecht um Google Drive zu benutzen.",
-    "gdriveReadOnly": "Diese Google Drive-Projekt ist schreibgeschützt.",
+    "drivFileeReadOnly": "Diese {drivename}-Projekt ist schreibgeschützt.",
```
neilccbrown, line 72: Typo in the key here.

neilccbrown (approved):
This was a lot!  Good job, must have taken a lot of effort.  Just a few small comments here and there to look at.

src/components/CloudDriveHandler.vue
```diff
@@ -0,0 +1,635 @@
...
+import { CustomEventTypes, getAppSimpleMsgDlgId, getFrameUID, getSaveAsProjectModalDlg } from "@/helpers/editor";
+import { pythonFileExtension, strypeFileExtension } from "@/helpers/common";
+import { BootstrapDlgSize, SaveRequestReason, StrypeSyncTarget } from "@/types/types";
+import { CloudDriveAPIState, CloudDriveComponent, CloudDriveFile, CloudFileSharingStatus, SaveExistingCloudProjectInfos } from "@/types/cloud-drive-types";
+import GoogleDriveComponent from "@/components/GoogleDriveComponent.vue";
+import OneDriveComponent from "@/components/OneDriveComponent.vue";
+import { generateSPYFileContent } from "@/helpers/load-save";
+import { AppSPYFullPrefix } from "@/main";
+
+// This enum is used for flaging the action taken when a request to save a file on Google Drive
```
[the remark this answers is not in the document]
PwtKCL: It is for all, I should have updated the comment. I'll address that - thanks for pointing it out.
(In fact, I should really do a round up check about "Google Drive" at some point. I'll keep that for later.)

src/components/CloudDriveHandler.vue
```diff
@@ -0,0 +1,635 @@
...
+            Actions, // this is required to be accessible in the template
+            saveExistingCloudProjectInfos: {} as SaveExistingCloudProjectInfos,
+        };
+    },
+
+    created() {
+        // We do not add the specific Cloud Drive's components via the <template> part, to allow us typing them properly
+        // (see the cloud types file, it's a struggle to work that out).
+        // Anyway, there is no styling require for these components so it's not a loss...
+        // Google Drive
```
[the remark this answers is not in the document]
PwtKCL: Actually I think the whole comment makes not much sense. It might have been during a work in progress that I wrote that and forgot to remove it. I'll scrap it off totally.

src/components/OneDriveComponent.vue
PwtKCL: Yes, that's correct: I've left some "TODO" for future implementation. So the comments will be updated later.

src/components/CloudDriveHandler.vue
```diff
@@ -0,0 +1,635 @@
...
+                            this.proceedFailedConnectionCheckOnSave(cloudTarget);
+                            return;
+                        }
+                        
+                        // Folder not found
+                        if(responseStatusCode == 404){
+                            this.appStore.strypeProjectLocation = undefined;
+                            this.appStore.strypeProjectLocationAlias = "";
+                            this.appStore.projectLastSaveDate = -1;
+                        }
```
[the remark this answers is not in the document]
PwtKCL: That's a good point. Those codes was particular to Google Drive at the time. I'll add a catch all for the other 400+ error status.

PR #587, opened 2025-10-09
Various bug fixes

neilccbrown:
This fixes various bugs that I experienced while preparing and doing the recent workshop.  One was that bug you found with the collision detection and say(), and the rest were to do with editing, loading and saving the project/function descriptions, where there were issues with quotes and newlines, and ctrl-A wasn't working in Firefox.   It ended up quite a few changes but although there's many commits, I think each commit makes sense by itself if you want to view it that way.


PwtKCL (approved):
The quotes are a nightmare when it comes with parsing and interpreting them... Well done for fixing that!
I think the PR is fine, but as there is one empty file in the changes, I would like to confirm with you if its change had not been included or if it should be deleted.
All good to merge otherwise.

PR #594, opened 2025-10-15
Add support for pasting SVG as image literals

### LINKED ISSUE #586: Add support for pasting SVG images by converting to bitmap
Workshop participants asked what images can be pasted in and I said everything except SVG, then tried to paste the KCL logo from our webpage as an example... and that is an SVG.  But it occurred to me after that we could also allow pasting SVG by converting to bitmap -- this should be possible in browser, see e.g. https://stackoverflow.com/questions/3975499/convert-svg-to-image-jpeg-png-etc-in-the-browser

neilccbrown:
This came about from a workshop where I tried to demo image pasting, and copied our KCL logo from the front page and it wouldn't paste.  Turns out (a) it's an SVG and (b) it's an img tag with an SVG src.  We did actually already support copying SVG files from the file system (by accident?!), but browser copying is another thing because the browser copies the SVG (or IMG tag) to the clipboard as text/html content.  So we need to spot that and then act accordingly to get the image content out as a PNG.

Fixes #586


PwtKCL (approved):
Good finding! It might well happen to users so it's good that is fixed.
I have a suggestion regarding the CORS issue (if applicable). 
But the code can be merged as is.

PR #597, opened 2025-10-20
Initial work on the frozen-frame feature

neilccbrown:
This adds the internal state for frames being frozen, the context menu item for doing so, and loading/saving frozen frames.  There's not yet the implementation of freezing (which changes editing and visibility behaviour) but I think this is enough for one PR then I'll do the next bits.  As part of this I've changed from how we saved Collapsed state to a more general FrameState which can have any number of attributes which is more flexible for the future.  This won't affect any existing code because Collapsed was only in the oop branch which hasn't been released yet.

We are calling this frozen rather than locked because it's not just about editing, but also visibility of the child items, and it's an operation that can be reversed without a password (locked suggests that someone else cannot immediately unlock -- but in our design they can).  So the icon is a snowflake rather than a padlock.


PwtKCL (commented):
It seems all good, the only thing I think needs clarification is about the selection of frames and how freeze can (or cannot) apply to the them.

PR #605, opened 2025-10-23
Changed "save as" behaviour and a fix

PwtKCL:
Some old and odd behaviour of the "Save as" mechanism was left in Strype: when saving as a new file, we also saved the original file we were working on (when it was a cloud based file).
Now we ignore the autosave call in that context to only save the "new" file.

Also, I made a small mistake with the mechanism in place to load a Cloud-based project the fist time (i.e. when we had never loaded a file from the Cloud before). 
In this scenario, we try to open the "Strype" folder if it exists. I was initially relying on the state having the location ID set, but in the context of multi clouds, we don't want to set the location ID in the state yet before the file is picked.
That meant we were opening the picker (Google Drive) on "all Strype files" rather than Strype because the picker saw no valid ID.
So instead, we pass on the resolve folder ID (that is if the folder was found, Strype or any other actually) down to the picker itself.
(It only concerns Google Drive, OneDrive uses the path rather than ID to set the starting folder).

_That should be the last PR within the OneDrive branch. The plan is that next I'll merge main in and do a PR for the merging OneDrive in the main branch._

### DIFF
--- a/src/components/CloudDriveHandler.vue
+++ b/src/components/CloudDriveHandler.vue
@@ -457,7 +457,7 @@ export default Vue.extend({
                 // Set the sync target 
                 this.appStore.syncTarget = cloudTarget;
                 this.appStore.isEditorContentModified = false;
-                // Reset the "Save As" flag of the Men
+                // Reset the "Save As" flag of the Menu
                 (this.$parent as InstanceType<typeof Menu>).requestSaveAs = false;
                 // Set the project name when we have made an explicit saving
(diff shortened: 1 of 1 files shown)


src/components/CloudDriveHandler.vue
```diff
@@ -457,6 +457,8 @@ export default Vue.extend({
                 // Set the sync target 
                 this.appStore.syncTarget = cloudTarget;
                 this.appStore.isEditorContentModified = false;
+                // Reset the "Save As" flag of the Men
```
neilccbrown: Small typo in the comments throughout this file: Men -> Menu

PR #606, opened 2025-10-27
Prevented deletion and editing of frozen frames

neilccbrown:
Continuing work on the freeze feature.  This prevents navigating via keyboard into the slots of a frozen frame, and prevents deleting the entire frame.  Also includes tests for these features.  Still some more work to do on this but getting there!

### DIFF
--- a/src/components/Frame.vue
+++ b/src/components/Frame.vue
@@ -982,7 +982,7 @@ export default Vue.extend({
         },
 
         getBodyMidFramePositions(): {caretPos: CurrentFrame, midYThreshold: number}[] {
-            if (this.appStore.frameObjects[this.frameId].frozenState == FrozenState.FROZEN) {
+            if ((this.appStore.frameObjects[this.frameId].collapsedState ?? CollapsedState.FULLY_VISIBLE) != CollapsedState.FULLY_VISIBLE) {
                 return [];
             }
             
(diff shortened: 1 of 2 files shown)


src/components/Frame.vue
```diff
@@ -982,6 +982,10 @@ export default Vue.extend({
         },
 
         getBodyMidFramePositions(): {caretPos: CurrentFrame, midYThreshold: number}[] {
+            if (this.appStore.frameObjects[this.frameId].frozenState == FrozenState.FROZEN) {
+                return [];
```
PwtKCL: I don't get this part of the code. It returns an empty array but shouldn't it return something?
If I'm not mistaken, this part of the code is used to know where to place the frame cursor when we click on a frame (above? below?). So even if a frame is frozen, I suppose clicking on it will place the frame cursor somewhere.

src/store/store.ts
```diff
@@ -632,6 +632,17 @@ export const useStore = defineStore("app", {
             return state.frameObjects[frameId].isVisible;
         },
 
+        isEffectivelyFrozen: (state) => (frameId: number) => {
+            while (frameId > 0) {
```
PwtKCL: Maybe had a small explanation comment here to indicate the notion of "effectively".

PR #613, opened 2025-10-29
Prevent editing of frozen slots

neilccbrown:
This is a small PR but it's a bit unrelated to the next changes I'm making.

Essentially, we need to stop editing of frozen slots so I've changed the contenteditable attribute of the LabelSlotsStructure spans to be based on whether the frame is frozen.  

Inevitably Firefox then had something odd: because the LabelSlot div is also contenteditable Firefox allowed focusing at the edges of the div besides the spans, so we also need to turn off contenteditable on there if it is frozen.  I'm a little less sure about the change in LabelSlot's contenteditable (should I just look at the frozen state?).  See what you think.

### DIFF
--- a/tests/playwright/e2e/load-save-frozen-collapsed.spec.ts
+++ b/tests/playwright/e2e/load-save-frozen-collapsed.spec.ts
@@ -318,7 +318,7 @@ def __init__ (self, ) :
 `);
     });
 
-    test.only("Freezing prevents focusing the text slots with clicking", async ({page}) => {
+    test("Freezing prevents focusing the text slots with clicking", async ({page}) => {
         await loadContent(page, testState({"Alpha": "Frozen", "__init__": "FoldToHeader"}));
         await page.locator("span", {hasText: "Alpha"}).click();
         await page.waitForTimeout(3000);


src/helpers/editor.ts
```diff
@@ -158,7 +158,7 @@ export function parseLabelSlotUID(UID: string): SlotCoreInfos {
     return res;
 }
 
-export function isElementLabelSlotInput(element: EventTarget | null): boolean{
+export function isElementLabelSlotInput(element: EventTarget | null): element is HTMLSpanElement {
```
PwtKCL: Never thought this can be used like that.
But then, is the test below still necessary?

src/components/LabelSlot.vue
```diff
@@ -23,7 +23,7 @@
             @keyup.backspace="onBackSpaceKeyUp"
             @keydown="onKeyDown($event)"
             @contentPastedInSlot="onCodePaste"
-            :class="{[scssVars.labelSlotInputClassName]: true, [scssVars.navigationPositionClassName]: isEditableSlot && !isFrozen, [scssVars.errorSlotClassName]: erroneous(), [getSpanTypeClass]: true, bold: isEmphasised, readonly: (isPythonExecuting || isDisabled)}"
+            :class="{[scssVars.labelSlotInputClassName]: true, [scssVars.navigationPositionClassName]: isCurrentlyEditableSlot, [scssVars.errorSlotClassName]: erroneous(), [getSpanTypeClass]: true, bold: isEmphasised, readonly: !isCurrentlyEditableSlot}"
```
PwtKCL: It seems fine. I would just make a round trip in the code the be sure we don't need to list *all* elements set to the class `scssVars.navigationPositionClassName` (i.e. even those that shouldn't be reached).
I don't think we need, but just to make sure.

Issue #611, opened 2025-10-27
Triple click doesn't select the whole text slot with Firefox


PwtKCL:
I haven't managed to understand why exactly this happens.
But it is clear that it comes from the selection we set ourselves because if the content of `App.handleDocumentSelectionChange() `is commented then the triple click works.

Issue #609, opened 2025-10-27
Weird text replacement behaviour in text slot


PwtKCL:
(Tested from a new project.)
I triple-clicked a text slot to select its content, then hit a key, for example t. 
I ended up with: 
<img width="168" height="44" alt="Image" src="https://github.com/user-attachments/assets/7d01176a-29a6-4c4a-9b7b-c9ae21577768" />

Issue #607, opened 2025-10-27
Navigation issue in text slots with start/end keys


PwtKCL:
When we have a long string in a text slot, the start/end keys do not behave as expected: it will go to the start/end of the whole text, rather than going to the start/end of the (visual) line - which is the expected behaviour.

PR #617, opened 2025-10-29
Various fixes (keyboard navigation)

### LINKED ISSUE #608: Selected text slot isn't removed when using "del" key
I've triple-clicked a text slot to select the text. Then hit the "del" key, but the deletion didn't happen (however, the selection disappeared).
(Seems this needs to be when we are in a new project.)

PwtKCL:
This PR mainly addresses some issues/new features about the keyboard navigation in the editor.

There is an issue with macOS: home/end are not dedicated key and the equivalent is ⌘ + left/right.
In the following we will mention "home" and "end" **as the _functional_ role**: that means the keys are indeed home and end _for windows only_ but ⌘ + left and or ⌘ + right _for macOS_.

To summarise:
- only home/end is handled by the browser for comments (as it was before), string literals (#607) and documentation slots (both new behaviour) -- ,
- home/end *without selection* now moves to the start/end of the whole label slots structure for other cases, instead of moving to the current structural level
- home/send *with selection* behaves as before (that is it will go to the start/end of the current structural level)
- home/end keeps the same behaviour for the frame cursor (going to the top/end of the current frames section),  but now if ctrl is hold together, we reach the very top/end of the whole editor (that not collapsed)
- page up/page down now acts as up/down when within a text slot (of any kind); "scrolls" up/down the frame cursor when the frame cursor is showing

For Chrome, this PR also contains a fix for the triple click slot content selection (#611 which fixes #608 and #609).
**Note that at the moment, for Firefox at least, the triple click doesn't work at all - this is mentioned in a separate issue.**

Finally, the format of the label slots' id has been slightly modified to use greater SlotType values, notably so that it properly reflects the SlotType values for comments (i.e. documentation slots) and media.

### DIFF
--- a/src/components/LabelSlot.vue
+++ b/src/components/LabelSlot.vue
@@ -1139,10 +1139,10 @@ export default Vue.extend({
             //(*) Note that in macOS things are a big tricky because there is no dedicated home/end keys.
             //    This is their mapping (simplified for one direction only) and natural behaviour and what we will do:
             //      ⌘ + Left --> "metaKey + ArrowLeft" --> start of line     --> we detect that and let the browser do its job
-            //      Fn + Left --> "Home"                --> start of document --> we won't get it here
+            //      Fn + Left --> "Home"                --> start of document --> we detect that and go to the start/end of the whole slot
             //    The selection with shift is treated according to the rule above. It seemed that {Fn + Left + Shift} worked
             //    but to make things consistent we'll just drop it. 
-            if((event.key == "Home" || event.key == "End" || event.key == "ArrowLeft" || event.key == "ArrowRight") && (this.frameType == AllFrameTypesIdentifier.comment || this.slotType == SlotType.comment || this.slotType == SlotType.string)){
+            if(((!isMacOSPlatform() && (event.key == "Home" || event.key == "End")) || event.key == "ArrowLeft" || event.key == "ArrowRight") && (this.frameType == AllFrameTypesIdentifier.comment || this.slotType == SlotType.comment || this.slotType == SlotType.string)){
                 return;
             }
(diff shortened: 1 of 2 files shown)


src/components/LabelSlot.vue
```diff
@@ -1129,40 +1131,65 @@ export default Vue.extend({
...
 
         handleFastUDNavKeys(event: KeyboardEvent){
-            // If we are in a comment, we let the browser handling the key events.
+            // If we are in a comment (e.g. frame or documentation slot) or in a string slot, we let the browser* handling the key events.
             // Otherwise, the following rules apply:
-            // Home/End move the text cursor to the start/end of the "block" unit of the code, for example the start of a slot bracketed structure
-            // PageUp/PageDown: do nothing
-            if(this.frameType == AllFrameTypesIdentifier.comment){
+            // Home/End* move the text cursor to the start/end of the whole label slots structure, or of the current slot structure level
+            //      (like a bracketed structure) if shit is also held (to avoid selecting code that spans across different levels)
```
neilccbrown: Small typo...

src/components/LabelSlot.vue
```diff
@@ -1129,40 +1131,65 @@ export default Vue.extend({
...
                     setDocumentSelection(newAnchorSlotCursorInfo, {slotInfos: newFocusSlotCoreInfo, cursorPos: newFocusCursorPos});
                     this.appStore.setSlotTextCursors(newAnchorSlotCursorInfo, {slotInfos: newFocusSlotCoreInfo, cursorPos: newFocusCursorPos});
+
                 });
                 event.preventDefault();
                 event.stopPropagation();
                 event.stopImmediatePropagation();
             }
+            else{
+                // The case of PageUp and PageDown
```
neilccbrown: Maybe just mention again here: this leaves the slot to a frame cursor.

PR #626, opened 2025-11-02
Fix several bugs

neilccbrown:
This is three main things, for the OOP branch:

 - I discovered delete at the top of class frames wasn't deleting the default constructor because of some logic originally there to prevent backspacing frames from trying to put the content into the top-level definitions.
 - I found that # in triple-quoted strings for project description fields wasn't parsing correctly because of our code for nested quotes and comments.  I've rewritten that code now, based on a bit of GPT suggestion for the algorithm.

As part of that last work I simplified slightly our triple quote and newline handling.  I transform all the newlines into an escape string so that all triple-quoted strings end up being on one line each.  That way we don't have to deal with indentation issues while scanning afterwards (where otherwise we had to know if that indent was in a triple quote or not).

### DIFF
--- a/src/helpers/pythonToFrames.ts
+++ b/src/helpers/pythonToFrames.ts
@@ -195,7 +195,7 @@ function parseWithSkulpt(codeLines: string[], mapErrorLineno : (lineno : number)
         parsed = Sk.parse("pasted_content.py", codeLines.join("\n"));
     }
     catch (e) {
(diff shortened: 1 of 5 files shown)


PwtKCL (approved):
Just some comments in the code that can be removed, otherwise all good for merging.
Sorry for the flaky triple quote handling previously written, it's tricky part of the code, well done for fixing things up with it

Issue #619, opened 2025-10-30
A frame with only space in the text slot can be deleted with backspace


PwtKCL:
Remove this functionality, a frame can only be deleted if truly empty.

PwtKCL:
That's making sense only for comments and frame containing string literals

PR #627, opened 2025-11-03
A few fixes to address some found bugs and change for the Strype Graphics API

PwtKCL:
Generally, the commits explains the bugs or change.

For the Strype Graphics, I've made the change to show the documentation/prompts without mention of "filename" for images, since we decided not to show that anymore. The functionality isn't removed in the internals.
Note that I've tried to be very conservative with the doc change, so that the changes only concern the point above and nothing else about the documentation for a/c in general.
Separately, there is a MR pending on GitLab for the API documentation update online.

For this commit "Do not delete frames that contains comments or only string literals with spaces only" (#619) I am not 100% sure if that was what we wanted to do.
Basically I change the code so that if a comment that contains X, or a frame that only contains a string literal with X, <X> is considered empty for deletion of the whole frame is X is of 0 length.
I think that's what we said, but I don't really see why we would have, in a Python code, such code at any time.. it's pointless..

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -1199,7 +1199,12 @@ export interface LoadedMedia {
     audioBuffer?: AudioBuffer,
 }
 
-export interface MediaDataAndDim {dataURI: string, width: number, height: number, itemType: string}
+export interface MediaDataAndDim {
+    dataURI: string,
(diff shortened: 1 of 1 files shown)


neilccbrown (commented):
Generally good, just two comments to look at.

PR #630, opened 2025-11-04
Fixed Ctrl-A in definitions with OOP

neilccbrown:
This is not quite as big as it looks as a lot of it is a new test file.  Essentially, I needed to adjust the Ctrl-A functionality so that it worked for classes, but it was a bit fiddly because functions can be inside classes so some of the previous logic wasn't quite right (e.g. if you're just below a function, you're not necessarily at top-level now because you could be inside a class).  After a few failed attempts I tweaked the enum and rewrote the code a bit.

### DIFF
--- a/src/components/Commands.vue
+++ b/src/components/Commands.vue
@@ -351,12 +351,12 @@ export default Vue.extend({
                     if(eventKeyLowCase === "a" && !isEditing){ 
                         if(getActiveContextMenu() == null && !this.appStore.isAppMenuOpened){
                             const frameContainerId = getFrameSectionIdFromFrameId(this.appStore.currentFrame.id);
-                            // If a selection already exists, we clear it, after checking where we are at in the case of function defs (cf. below)
+                            // If a selection already exists, we clear it, after checking where we are at with respect to the select-all action to take
                             const selectAllFramesAction = getCurrentFrameSelectAllAction();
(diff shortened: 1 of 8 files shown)


src/components/Commands.vue
```diff
@@ -365,36 +364,33 @@ export default Vue.extend({
...
-                                        //then we don't need to look up for the body position as we're already there...
-                                        foundFunctionDefFrame = true;
-                                        functionDefFrameId = currentFrameId;
-                                    }
-                                    while(!foundFunctionDefFrame){
-                                        functionDefFrameId = this.appStore.frameObjects[currentFrameId].parentId;
-                                        foundFunctionDefFrame = (this.appStore.frameObjects[functionDefFrameId].frameType.type == AllFrameTypesIdentifier.funcdef);
-                                        currentFrameId = functionDefFrameId;
+                            case SelectAllFramesAction.functionOrClassContents:
+                            case SelectAllFramesAction.currentLevel: {
```
PwtKCL: I think it was good to have a small comment here to explain the cases we're in

PwtKCL (commented):
Good call thinking about the select all case... and rewriting it properly.
I have only some commenting suggestions otherwise all good.

PwtKCL (approved):

PR #634, opened 2025-11-05
Fix how group toggling of collapsed state works

neilccbrown:
I would only view the final changes because I had a few detours along the way.  This fixes the group toggling of collapsed/folded state to work as agreed in last meeting.  It also makes ctrl-clicking of visibility controls on a frame do the same toggling to the whole selection.  My notes on how it should work:

 - We have a memory of the toggle; mixed state goes to folded, then goes to just doc then to normal.  If anything changes outside the group toggling then the memory part is ignored and you go back to fully folded then start remembering from there.


PwtKCL (commented):
That's a smart mechanism in place! 
It seems all fine, I've had only a Vue related suggestion (not sure it's worth doing, you can merge if you think it's not or want to do it later) and comment-related notes, otherwise all good.

PR #665, opened 2025-11-16
Prevent folding frames which contain syntax errors 

neilccbrown:
This implements the rule discussed by email, that frames with syntax errors in them should not be able to fold.

(I'm on leave Mon-Wed this week so I'll fix any comments on Thursday.)


PwtKCL (commented):
Just one comment for left "console.log". I've given you a French translation in one review comment. All good the merge otherwise.

PR #677, opened 2025-11-24
Various fixes of the issues

PwtKCL:
There is not a lot of bug fixed in this PR, but two of them are pretty severe and I think will need to be tested as soon as we can.
The bugs are mentioned in the commits.
The bug I think we will need to test the most is the one fixed by the last commit: the error showing at the wrong frames related to the parsing of the comments and documentation slots.


src/store/store.ts
```diff
@@ -2974,6 +2974,8 @@ export const useStore = defineStore("app", {
         },
 
         copyFrame(frameId: number) {
+            // We do not use the system's clipboard for frames, so we clear any potential text to avoid interference
+            navigator.clipboard.writeText("");
```
neilccbrown: It's odd: I was thinking this wouldn't make a difference, because we prefer frame pasting to text (i.e. if there are copied frames we use those, otherwise check the clipboard).  I was quite confident about this because I remembered I made an ill-fated change to put the text version on the clipboard and it caused problems.  But I think I backed out the putting the text on the clipboard during *copy*, not the preferring of text when *pasting*.  So this is a good fix, I guess this may have caused some of the copy/paste issues we saw?  I guess the problem would have been caused by copying frames with text on the clipboard already, then pasting and we would try to paste the text you had before, not the frames?

PR #681, opened 2025-11-26
Various fixes again and small additions to Strype

PwtKCL:
The fixes of this PR are:
- reduced the stack of undo/redo from 200 to 50 (I hope it's fine, because the stack _includes every input changes_ "letter by letter", which means the stack can actually be full pretty fast
- deal with problematic scrolling into view when we change the editor's focus. _As explained in the related issue ticket, that's not ideal but it is better than before (I hope!)._
- deal with getting the focus out the project documentation slot by a click in a void area

Additions/enhancements are:
- change the "new project" mechanism (using a modal dialog we show ourselves) so we limit the flaky outcomes due to the timeout we used before)
- disable the Strype menu (and related actions) when the project is running (and show a highlight on "stop")
- updated the default project/examples _(note that Michael told me what he wanted to see for the default project, but the examples' project documentation were still to be discussed. We can always amend them later...)_
- added "myClasses" in a/c to show up the classes now available with the class frame. _Note that t**here is still one thing to add** for that: the support for a/c about "self" inside the classes. I've left it out for the time being, so the related issue should still be opened._
- change the a/c showing suggestions that _starts with_ the token to show suggestions that _contains_ the token, _That makes the a/c suggestions less restrictive, but I think it's essential for young users who may just remember a word from a method etc._

### DIFF
--- a/tests/playwright/e2e/structured-expressions-navigation.spec.ts
+++ b/tests/playwright/e2e/structured-expressions-navigation.spec.ts
@@ -110,10 +110,10 @@ test.describe("Check navigation", () => {
         for (let i = 0; i < 100; i++) {
             await checkFrameXorTextCursor(page);
             await page.keyboard.press("ArrowDown");
-            await page.waitForTimeout(200);
+            await page.waitForTimeout(500);
             await checkFrameXorTextCursor(page);
             await page.keyboard.press("ArrowDown");
-            await page.waitForTimeout(200);
+            await page.waitForTimeout(500);
             await checkFrameXorTextCursor(page);
             await page.keyboard.press("Shift+Tab");
         }


public/demos/graphics/particles.spy
```diff
@@ -2,6 +2,7 @@
 #(=> peaLayoutMode:tabsCollapsed
 #(=> peaCommandsSplitterPane2Size:{"2":68.42,"tabsCollapsed":48.22}
 #(=> peaSplitViewSplitterPane1Size:{"2":77.23}
+'''A particule sparkling effect that follows the mouse'''
```
neilccbrown: Should be particle, in English.

neilccbrown (commented):
Generally fine so far, I can check more commits tomorrow if you need them for test fixing.

tests/cypress/e2e/load-save.cy.ts
```diff
@@ -218,7 +222,7 @@ describe("Tests blanks", () => {
 describe("Tests invalid characters", () => {
     it("Outputs a file with invalid chars", () => {
         testEntryDisableAndSave("{uparrow}{uparrow}" +
-            "i100{rightarrow}ffoo{rightarrow}£1000{downarrow}i50{downarrow}ifoo（）{downarrow}{downarrow}" +
+            "i100{rightarrow}ffoo{rightarrow}£1000{downarrow}i50{downarrow}if#(=>oo（）{downarrow}{downarrow}" +
```
neilccbrown: I don't quite follow how here we seem to have added four characters, but in the project-invalid-chars it only looks like one?

tests/cypress/fixtures/project-invalid-chars.spy
```diff
@@ -3,7 +3,7 @@
 import ___strype_invalid_u0031u0030u0030 
 from foo import ___strype_invalid_u00a3u0031u0030u0030u0030 
 import ___strype_invalid_u0035u0030 
-import ___strype_invalid_u0066u006fu006fuff08uff09 
+import ___strype_invalid_u0066u0023u006fu006fuff08uff09
```
neilccbrown: See earlier comment; this only looks like one more unicode char?

tests/cypress/e2e/load-save.cy.ts
PwtKCL: Sorry it's my mistake, I've added them accidently.

PR #687, opened 2025-11-28
Fixes to several bugs, especially with multiline comments/descriptions.

neilccbrown:
This is the other active branch I had (will return to dividers next).  Fixes two bugs: the decimal numbers as keys in a dict, and then issues with escaping/unescaping quotes in description fields, plus a bug I found where disabled frames were not being treated correctly when we had multiline comments/descriptions.


PwtKCL (commented):
There is only a console.log that may be removed - I'll merge the code for now, if that console.log should go it can be removed in your next PR.

PR #692, opened 2025-12-01
Save default parameters correctly

neilccbrown:
This fixes the bug where saving e.g.

```
def foo(a, x =0, y = 0):
    return 0
```

Was saving the params as invalid, because we had notated them as only accepting names and commas.  Now it spots the equals and understands which bits are allowed parameters.


PwtKCL (commented):
I just have a tiny comment about the test name, otherwise all good to merge.

PR #696, opened 2025-12-03
Add ability to share (and load) snapshot links

### LINKED ISSUE #693: Allow sharing of snapshot link as a long URL
We should offer an alternative way to share if you aren't saving to cloud, by giving a link that encodes the program state into the URL.  (We already have the load functionality, but we need the GUI for doing the original share.)

neilccbrown:
As discussed a little while back, this allows you to load snapshot links where the full code is in the URL itself.  We basically zip then base64 encode the SPY, and reverse the process on load.  If the URL is big (most browsers support 64K URLs, but some only 2K) then we warn about it.  Essentially: most normal programs will fit in 2K, but anything with image/sound literal is likely not to work.  But you can still do the cloud share in those cases.

The dialog looks like this:

image

I'd only look at the final changes as I had one or two false starts on the way.  Also includes tests, some in Cypress about loading the links and then some in Playwright for saving and loading, after I remembered Cypress doesn't support testing the clipboard.

Fixes #693


src/App.vue
```diff
@@ -724,6 +726,28 @@ export default Vue.extend({
...
+            if (m) {
+                const param = m[1];
+                // Base64 uses A-Za-z0-9+/, but the latter two are meaningful in URLs
+                // so we have to replace them (with - and _ respectively).
+                // That is handled automatically on decode by the js-base64 package: 
+                const binary = Base64.toUint8Array(param);
+                const spyContent = inflateRaw(binary, { to: "string" });
+
+                const loadSpy = () => this.setStateFromPythonFile(spyContent, this.$i18n.t("en.defaultProjName") as string, 0, false);
+
```
PwtKCL: Not entirely clear about the localisation key here. Don't we want to have the project name in the current locale? (so just having `defaultProjName`)

PwtKCL (commented):
I have one comment about the localisation, otherwise it's fine.
One test still fails, but I think it's not related to your PR.
I've relaunched it, if it passes then good to merge after the potential change for the comment mentioned above.

PR #698, opened 2025-12-04
Fix image literal bugs

### LINKED ISSUE #656: Reuse of image copy wrong behaviour
Observed with Safari (at least, investigate deeper).
The copy of an image pasted as frame, then reused in `Actor()` makes 2 images; trying to paste again elsewhere adds images to the first added frame.

### LINKED ISSUE #661: Wrapping a media thumbnail generates an issue
In Strype, use an image to make a frame like this (either create a frame, then paste, or paste at caret):
Image

Then select this and wrap it with brackets, the result is:
Image

Note that if I insert the brackets _before_ the image, than push the bracket with alt+right, it works fine.

### LINKED ISSUE #663: Copy/Paste of image slot not working?
Select only a media slot (image - tested with forward selection), then copy, then paste somewhere else: it doesn't work - the copy seemed to be ignored as I had what I thought would be the previous element of the clipboard.
However, if you select more than just the media slot, it works.

neilccbrown:
This fixes several image literal bugs: wrapping them in brackets, pasting them at the frame cursor, and copying them from within slots (by themselves) to paste again.  The individual commits go a bit back and forth and cross-over so I'd look at the end result.  Main changes:

 - We still handle keyup and keydown (things broke when I removed the keyup part) but the ctrl-a and ctrl-x/c only trigger on keydown.
 - We do the pasting at frame cursor slightly differently with timeouts.  Part of the issue here was an exception which was occurring when looking for cursor positions before/after the refactoring so I took a slightly different approach.
 - When copying images from slots we only copy the image rather than image and text, which makes things simpler.

Fixes #663 
Fixes #661 
Fixes #656


PwtKCL (approved):
All seems good. Even if tests are currently failing, we can merge this PR because the tests fails aren't related.

PR #707, opened 2025-12-08
Various small improvements

neilccbrown:
This fixes various small things, several requested by Michael:
 - Changed the docs generation to remove the package prefix in each name (so set_background, instead of strype.graphics.set_background), and to compensate, added a "sticky" item on the right showing the package.
 - Prevented line break in functions between the bracket and colon, and fixed problems that introduced.  (Now that I write this I wonder if I should have just removed that space instead!?  What do you think?)
 - Made the folding/freezing controls be top-aligned.  Previously, if the function header was too long and wrapped, the folding control was vertically centred against the multiple lines of the function header, which looked odd.
 - If image and text on clipboard, only paste image
 - When we paste Python text at frame cursor, go through and update the parameter prompts.  Since this is calculated in the refactoring check, I actually issue a refactoring check on all slots in all newly added frames.
 - Fix one issue that turned up which is that if you had "try... except Exception as e:" then the refactoring was turning it into "Exceptionase" (Sounded French or Italian when I first saw it!) because it didn't know that as is valid operator in except.  (Previously it was only valid in import).

The PR is not as long as that sounds!  Each one is only a few lines of code change.

### DIFF
--- a/docs-sphinx/source/_static/custom.css
+++ b/docs-sphinx/source/_static/custom.css
@@ -58,6 +58,7 @@ h2 {
     position: relative;
     padding-bottom: 1rem;
     border-bottom: 1px solid rgba(0,0,0,0.05);
+    margin-bottom: 150px; /* Good to have space anyway, and stops the scrolling indicators overlapping */ 
 }
 
 /* The indicator itself */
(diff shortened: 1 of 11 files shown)


PwtKCL (approved):
Good. To reply the question asked in the PR description: not having `&nbsp;` for the function definition label would result in an add spacing in the UI. So we can keep it.

PR #708, opened 2025-12-09
Fix saving/loading of folding and freezing

neilccbrown:
This is the same fix as for disabled frames when dealing with multi-line comments.  If you expand the file slightly you'll see the changed line on 377 now matches pythonToFrames line 357.  Also added some tests to show the problem.


PwtKCL (approved):
Good to merge when tests pass

Issue #713, opened 2025-12-23
Errors persist on description slots


neilccbrown:
Make a class with a method and fill it in like this and leave the header slots:

<img width="598" height="236" alt="Image" src="https://github.com/user-attachments/assets/76880445-7e6b-4371-bdb0-3b3b94f05305" />

An error appears in the description slot because TigerPython is confused by the whole syntax arrangement.

Then go back and correct it and leave again:

<img width="538" height="252" alt="Image" src="https://github.com/user-attachments/assets/41c39278-b87d-47fe-b285-b714c657942d" />

Note how the error is gone from the first line of the header but the errors on the function description slot remain.  It is not actually possible to clear these errors, even if you edit the description or the slots before it.

We should either never put the errors on the description slots, or should clear them when any slot in the header is edited.

Issue #712, opened 2025-12-21
Slot errors are not shown immediately


neilccbrown:
Make a new Strype project.  Make a new assignment frame, enter `x = 1 *` then press right arrow to leave slot.  You will see this:

<img width="540" height="292" alt="Image" src="https://github.com/user-attachments/assets/4836d821-b66e-4419-ba3e-69e9ee786759" />

No error is shown in the blank slot after the multiply operator.  If you focus and defocus a text slot (can be an unrelated one, like the slots with `myString` below), it is then shown.  This is wrong; it should be shown when you first leave the assignment frame.

Issue #629, opened 2025-11-03
Images slow down the UI


PwtKCL:
I've realised that having images inside a project impacted quite a lot the speed of the text input with Strype.

I think it goes along the line of parsing the code or something similar - but I have not great idea how we could sort things out.

PR #714, opened 2025-12-23
Fix various problems with slots and errors

neilccbrown:
(Not urgent, can wait until next year, but filing it while I remember what I did...)

I set out to speed up slots with media literals, which I eventually did (we basically blank the strings in load_image/load_sound calls when generating code to feed to TigerPython, which seems to lex long strings fairly slowly).  This makes it noticeably faster as we feed code to TigerPython often (for autocomplete on each edit, and for errors on most edits and run).

But I first wrote some tests to check that it didn't upset error positioning (we already have autocomplete tests).  And I discovered two bugs which I filed and fixed: errors were not shown immediately when defocusing a newly created frame, and errors could be shown on function description fields in some cases (where TigerPython wanted to position them around/after the colon of the function definition).

And as a final bonus commit I happened to notice our grey/gray colour names were inconsistent, so I fixed them while I was doing a PR anyway.  Individual commits and the bug reports (#629, #712, #713) have more details


tests/playwright/e2e/slot-errors.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    // I can't figure out why.  I've tried them manually in Chrome and Chromium on the same
+    // machine and it works fine, but I see in the video that the test fails in Playwright
+    // (pressing right out of a comment frame puts the cursor at the beginning and makes a frame cursor).
+    // Since it works in the real browsers, and on Webkit and Firefox, we just skip the tests in Chromium
+    test.skip(testInfo.project.name == "chromium", "Cannot run in Chromium");
+
+    await page.goto("./", {waitUntil: "load"});
+    await page.waitForSelector("body");
+    //scssVars = await page.evaluate(() => (window as any)["StrypeSCSSVarsGlobals"]);
+    //strypeElIds = await page.evaluate(() => (window as any)["StrypeHTMLELementsIDsGlobals"]);
```
PwtKCL: The comments can be removed

tests/playwright/e2e/slot-errors.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+        await checkErrorAfterExitingSlot(page);
+    });
+    test("Empty subscript", async ({page}) => {
+        // Assignment, x = 1 * <err>
+        await page.keyboard.type("=x=a[");
+        await checkErrorAfterExitingSlot(page, ["ArrowRight", "ArrowRight"]);
+    });
+    test("Empty subscript (exit with arrow down)", async ({page}) => {
+        // Assignment, x = 1 * <err>
+        await page.keyboard.type("=x=a[");
```
PwtKCL: Same thing about updating the comment to the test

PwtKCL (commented):
All good code-wise, I've only mentioned a few commenting things that can be either fixed in another iteration or in another commit for this PR. It can be merged.

neilccbrown:
All comments addressed, will merge.

Issue #716, opened 2026-01-06
Slot background flickers during editing


neilccbrown:
This is especially visible on Firefox (which I think maybe repaints at different times?), but if you edit a slot you'll see that the background colour flickers between white (focused) and coloured (unfocused).  This is because in the refactor-checking code we lose the focused: attribute on the relevant slot (even if no real refactor change was made) and it is not restored until nextTick.  It seems that Firefox (at least) renders in this period and hence you see the flicker as the slot goes from focused: true to no focused attribute and back to focused: true again.  We need to try to retain the focused attribute.

See video for flicker:

https://github.com/user-attachments/assets/4307bdec-cb92-4aa0-b587-fab701d102ea

PR #718, opened 2026-01-06
Several small fixes

neilccbrown:
The fixes are small in that the code is short, but two of them in particular are quite wide-reaching and may need some testing:
 - A fix for a flickering focus issue (see #716) by trying to keep the focused flag set while refactoring slots.  It seems to work from my testing but if we get it wrong it could lead to slots having a different colour until focused/unfocused again.
 - A fix for an issue Michael reported where the caret would be scrolled to when we didn't want to.  One example was right-clicking the top of a very long frame (e.g. a class); we would select and scroll all the way to the bottom of the class to see the frame cursor.  Another was if your frame cursor was in a long list of frames and you used the multi-fold toggle to toggle them all out, you would scroll down to see the frame cursor, and have to scroll back up to toggle again.  The scroll-into-view was happening whenever the frame cursor changed; I've changed it to scroll only on an explicit event which is omitted in the (relatively few) places in the code where we change the frame cursor.

### DIFF
--- a/docs-sphinx/source/_static/custom.css
+++ b/docs-sphinx/source/_static/custom.css
@@ -93,3 +93,12 @@ h2 {
     border: 1px solid rgba(0,0,0,0.1);
     border-radius: 4px;
 }
+
+/* Little bit fragile, but add space before all modules after first */
+.sphinxsidebarwrapper > div > ul > li> ul > li:not(:first-child) {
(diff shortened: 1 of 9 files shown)


PwtKCL (approved):
Just to make things clear, I'll leave this PR unmerged so it can be merged once the previous PR has been merged.

PR #733, opened 2026-01-14
Fix up/down in description fields (especially project comments)

neilccbrown:
Project comments weren't working because their frame header lacked an ID so that was a simple fix.  But it turns out up/down wasn't working right at the start of lines because of the way browsers do bounds for zero-sized selections (like a cursor usually is).  Essentially if you have this:
```
one
two
three
```
which as a string in a <span> is "one\ntwo\nthree" then if you have the cursor position after a \n, like "one\n|two\nthree" then all the browsers will tell you that cursor position is at the end of the first line, not the beginning of the second.  The work-around is to ask for the width of the character "t" at the start of two and then it works.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -50,6 +50,7 @@
                                     @mousedown="handleWholeEditorMouseDown"
                                 >
                                     <FrameHeader
+                                        :id="getFrameHeaderUID(-10)"
                                         :labels="projectDocLabels"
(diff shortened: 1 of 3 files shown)


PwtKCL (commented):
It seems a change is required for the tests as one fails. Otherwise, all good to merge when tests are amended (if that's the problem indeed).

PR #737, opened 2026-01-20
Added two new graphics methods: get_actor_at and show_text

### LINKED ISSUE #734: Add a showText function
Greenfoot has a showText(x, y, text) function with the semantics that text is shown at (integers) x, y.  If you want to remove it, call with same x, y and the empty string.  That worked quite well, and we should add same to Strype.  Maybe with optional params like on say() to control sizing and max width and font family.

### LINKED ISSUE #735: Add a function get_object_at
We agreed to add a top-level function get_object_at(x, y) which gets the Actor touching location x, y.  If many objects overlap with that point, give back an arbitrary pick.  Maybe we should also allow an optional tag parameter?

neilccbrown:
As mentioned in the meeting, a small change to add two new methods (and fix a bug I found in draw_text) and a few tests for them.

Fixes #734 
Fixes #735


PwtKCL (commented):
Just a small comment suggestion for the doc.
Out of this MR, I wonder if we should have a small API doc version showing up somewhere (internally for us) to know when we've updated the doc in the python file but not on the server.
I've reloaded the test as I think it will pass with a reload. Either you or me can merge this MR when we notice the tests passed.

PR #738, opened 2026-01-20
Media improvements

### LINKED ISSUE #480: Image literal editor: add a rotate/mirror?
If users copy and paste images inside `Actor()`, for using Graphics, the image they had may not be exactly orientated the way they want to use it in their code.
Should the image edition tool also allow rotating and/or mirroring the image?

### LINKED ISSUE #491: Add a download button for media literals
Add a download icon button for media literals in the hover preview.

neilccbrown:
So I put these off because I thought they would be tricky and they both turned out to be very simple...  Adds a download button to the media preview popup, and flip/rotate buttons to the image editing dialog (which it turns out has built-in support in the cropper component we are using).

Fixes #480 
Fixes #491


PwtKCL (approved):
All good. I'll add the French translations when I rebase my work on yours.

PR #749, opened 2026-02-09
Add support for strype.graphics and strype.sound libraries under Pyodide

neilccbrown:
This PR changes all of the graphics, sound and input Strype code to use Pyodide rather than Skulpt.  Since Pyodide runs in a web worker, this involved quite a lot of new architecture for communication between the web worker and the main thread.  Sorry that it's so big but I changed the architecture multiple times along the way so ultimately I think it was best this way (definitely only review final version, going commit-by-commit is very confusing on this).

Some notes for review:

 - I've put a long comment at the top of python-execution.ts which I'd suggest reading first to understand things overall, plus also then the comment at the top of worker_bridge_type.ts.  I'd definitely recommended jumping to those and reading those comments first, especially python-execution.ts.  After that the code should make a bit more sense.
 - Most of the changes in the graphics.py and sound.py are very small due to a rename (I added this name "Sprite" as the old name "PersistentImage" was not good) or small changes in how Pyodide/Skulpt pass data through.  The real changes are in the internal Javascript modules (which have now become Typescript thanks to Pyodide/Vite).  Unfortunately Git sees these files as having been fully deleted and fully re-added (because it changed path and name and file extension), but that's maybe not the worst as the diff would have been very choppy.
 - Ignore the pysrc/python_runner/* code, that is pulled directly from https://github.com/alexmojaki/python_runner but we need the full source because we need to combine it into a single zip with our strype.* code as Pyodide wants all "built-in" Python code in one zip file.
 - I changed back how the acManager.ts imports the graphics and sound files as the fetch method doesn't work now that the files no longer live inside public.  I saw your note but I didn't see any warnings but happy to look into fixing them if they are still there.
 - Your commit to fix the cyclical import by adding helpers/appContext is in the PR but it's so small it shouldn't really confuse anything (and you know that code already).
 - Beware that a couple of relevant files (PythonExecutionArea.vue and image_and_collisions.ts) have been collapsed by default.

I'll double-check the tests are doing what I expect.  There will be failures, as I haven't done turtle at all yet, and I also need to fix some bits to do with when we preview image literals in the main canvas.  And autocomplete for built-in items (e.g. basic functions like len) needs looking at again, I think we might be able to make that simpler with Pyodide.  But at the moment the strype.* Python modules should all work and do the right thing.

### DIFF
--- a/pysrc/strype/graphics.py
+++ b/pysrc/strype/graphics.py
@@ -299,8 +299,8 @@ class Image:
             loaded = _strype_graphics_internal.canvas_loadFont("google", font_family)
             if not loaded:
                 raise Exception("Could not load font " + font_family)
-        dim = _strype_graphics_internal.canvas_drawText(self.__image, text, x, y, font_size, max_width, max_height, font_family).to_py()
-        return _Dimension(dim['width'], dim['height'])
+        dim = _strype_graphics_internal.canvas_drawText(self.__image, text, x, y, font_size, max_width, max_height, font_family)
+        return _Dimension(dim[0], dim[1])
         
     def draw_rounded_rect(self, x, y, width, height, corner_size = 10):
         # type: (float, float, float, float, float) -> None


src/stryperuntime/worker_bridge_type.ts
```diff
@@ -0,0 +1,226 @@
...
+    | { request: "canvas_downloadPNG", img: RemoteCanvas, filenameStem: string }
+    | { request: "startSound"; sound: RemoteSound }
+    | { request: "stopSound"; sound: RemoteSound }
+    | { request: "setMonoSoundSampleValues"; sound: RemoteSound; values: number[]; targetOffset: number }
+    | { request: "downloadWAV"; sound: RemoteSound; filenameStem: string }
+;
+
+// These are not used so we need to disable warnings, but they give errors if we've messed up above and tried
+// to send a type we are not allowed to send:
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
PwtKCL: I don't understand the comment clearly: "but they give errors if we've messed up above and tried
// to send a type we are not allowed to send:" --> do you mean "the request calls"  for "they", based on your other comments?
Otherwise, I get the idea why you do it, as it's explained in the file preamble.

src/helpers/execPythonCode.ts
```diff
@@ -154,6 +154,7 @@ export function handleErrorTrace(errorType : string, traceback: { filename: stri
     let errorLine = -1;
     if (traceback) {
         let lastmodule = "";
+        console.log("Traceback", Array.from(traceback.entries()).map((e) => e[1].filename + "@" + e[1].lineno));
```
PwtKCL: I suppose you want to keep it for some time for the continuing work? Maybe if so, add a TODO to remember to remove?

PR #751, opened 2026-02-20
Pyodide Autocomplete

neilccbrown:
Previously, we had an awkward system whereby we would ask Skulpt for its available items, then ask the local Python installation for documentation.  The new version still does the preprocessing step (since we only need to work out all the available items once, if the Python version doesn't change) but it entirely uses Pyodide for the query and for fetching the docs.  The old add-docs Python file has become one that fetches all the info but running inside Pyodide, not running locally.

The diff doesn't realise but process-skulpt-api.ts has become process-pyodide-api.ts (it thinks it was a delete and re-add).

I also needed to adjust a few tests because some of the functions we used (like urlopen) have changed their signature in the Python version that Pyodide uses, compared to the old one Skulpt was using.  Now all tests are passing except those involving Turtle, which is still an item for the future.

I'd look only at the final version as the commits along the way meandered a bit before I got it right.


scripts/process-pyodide-api.ts
```diff
@@ -0,0 +1,265 @@
...
+            sortedData[key] = [...data[key]].sort((a, b) => a.acResult < b.acResult ? -1 : a.acResult > b.acResult ? 1 : 0);
+        });
+
+    return sortedData;
+}
+
+
+const builtinModules = Object.keys(pythonBuiltins);
+const allContent : AcResultsWithCategory = library ? {} : {"": builtinModules.map((n) => n.endsWith("_$rw$") ? n.replace("_$rw$", ""): n).filter((func) => !func.includes("$")).map((s) => ({acResult: s, documentation: "", type: [], version: 0}))};
+// Add Skulpt's builtin functions to the default module:
```
PwtKCL: And here too.

PwtKCL (commented):
Just minor commenting to look at, but otherwise all seems good to merge in :)

PR #755, opened 2026-02-26
Vue3 migration

PwtKCL:
This PR contains changes on our code base to bring up Vue to the version 3.
The migration process isn't yet complete but it is at a stage where it is already big (sorry for this), and can do with a code review.
I've tried to squash commits by "changes target", but there may be overlaps that all come to the expected code at the end.
So I think you can check them one by one, but don't be surprised if you see some changes that seem a bit off topic in them....

In terms of migration:
- I've moved to Vue 3 and still use the migration compatibility package - the warnings it provides aren't yet fixed, I'll do that in my next phase.
- Most packages we use are upgraded to a newer version or a different package so that it becomes compatible with Vue 3. There are still a few ones I need to migrate as I expect them to stop working once we get rid of the compatibility package, but the main ones are migrated (like for Vue Bootstrap, SplitPanes..).

_You will see there is a file in the codebase called components.d.ts. This file is used for development (for TS typing) and is produced by the plugin "unplugin-vue-components" which is recommended by Vue Bootstrap Next._ 

In order to minimise the changes, I've only used the Composition API style (of Vue 3) in the Vue components when it is required. Our code was written using the Options API (Vue 2) which is supported by Vue 3.

### DIFF
--- a/src/components/Frame.vue
+++ b/src/components/Frame.vue
@@ -103,7 +103,7 @@ import JointFrames from "@/components/JointFrames.vue";
 import { DefaultFramesDefinition, CaretPosition, CollapsedState, CurrentFrame, FrozenState, NavigationPosition, AllFrameTypesIdentifier, Position, PythonExecRunningState, FrameContextMenuActionName, ContainerTypesIdentifiers } from "@/types/types";
 import VueContext, {VueContextConstructor}  from "vue-context";
 import { getAboveFrameCaretPosition, getAllChildrenAndJointFramesIds, getLastSibling, getNextSibling, getOutmostDisabledAncestorFrameId, getParentId, getParentOrJointParent, isFramePartOfJointStructure, isLastInParent, frameOrChildHasErrors, calculateNextCollapseState } from "@/helpers/storeMethods";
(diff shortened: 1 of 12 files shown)


src/components/Frame.vue
```diff
@@ -898,10 +932,8 @@ export default Vue.extend({
                 // However, since no actual slot is clicked, the change from "self" to "self," isn't triggered.
                 // We can retrieve the LabelSlotsStructure component because its ref is in the root object, and 
                 // call updatePrependText() which will now notice the right context and do its work.
-                const labelSlotsStructComponent = this.$root.$refs[getFrameLabelSlotsStructureUID(this.frameId, 1)];
-                if(labelSlotsStructComponent){
-                    (labelSlotsStructComponent as InstanceType<typeof LabelSlotsStructureComponent>).updatePrependText();
-                }
+                vueComponentsAPIHandler.labelSlotsStructureComponentAPI?.forInstance[getFrameLabelSlotsStructureUID(this.frameId, 1)]
+                    .updatePrependText();
```
neilccbrown: Does this last dot also need to be a "?."  ?  I suppose if it's missing something is wrong elsewhere...

src/vue3-burger-menu.ts
```diff
@@ -0,0 +1 @@
+declare module "vue3-burger-menu";
```
neilccbrown: I don't quite understand what this file is doing.  (I know it's not really changed from before but I didn't understand that one, either!)  Maybe a one line comment to explain?

PR #759, opened 2026-03-06
Vue3 remove compatibilty package

PwtKCL:
This is another intermediate PR. It is of decent size to have one now, and that should be the last about the work on updating packages for Vue 3 as such, before I address the core Vue compatibility issues (things that are still supported with the compatibility package and need to be addressed for a full Vue 3 version).

So in this PR, I've removed a few unused packages from the package.json and most importantly updated 2 packages that would not be compatible with Vue 3: the package we used for _the context menu_ (on frames,, caret and PEA) and the package we used for showing _the folder picker_ for saving to OneDrive against a work account.

Note that the change of package for the context menu simplifies our code somehow: first because it handles the "enter" key natively to select a menu entry, and second because we can no longer have 2 context menus showing at the same time. There is only one menu showing for the App. However, because the internal HTML structure is different, a few "magic" workarounds are made to keep the behaviour of the previous menu regarding hovering and keyboard selection.

### DIFF
--- a/src/components/ContextMenu.vue
+++ b/src/components/ContextMenu.vue
@@ -30,6 +30,8 @@ const props = defineProps<{
 watch(
     () => props.showContextMenu,
     (newVal, _) => {
+        // We only need to handle the case when the menu shows, as the Vue 3 context menu component already handles
+        // what we need to be removed when the menu is disused.
         if(newVal){
             showContextMenu();            
         }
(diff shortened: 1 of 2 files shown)


src/components/ContextMenu.vue
```diff
@@ -0,0 +1,183 @@
...
+    doAfterShownOnMenuItemKeyboardFocusRemovedObs?: (ctxMenuItemDef: StrypeContextMenuItem) => void,
+}>();
+
+// Watch when the showContextMenu property is set to true, to show the context menu
+watch(
+    () => props.showContextMenu,
+    (newVal, _) => {
+        if(newVal){
+            showContextMenu();            
+        }
```
neilccbrown: Does it also need the reverse, hide if turned to false?  If not, might be worth adding a comment as to why that's not needed.

PR #760, opened 2026-03-09
Pyodide cloud file support

neilccbrown:
This adds cloud file I/O support for Pyodide, by adapting the existing Skulpt one.  Also includes one or two small Pyodide fixes I noticed while developing: error messages were reported in the wrong place and it's best to restart Pyodide each run, even if it finished successfully.

(Note that eventually I'd like to rename the file skulptFileIO and remove the Skulpt bits, but I'd prefer to do that after merging in the Vue 3 changes otherwise I think it's going to make the merge much harder.)

As previously mentioned, the Pyodide file support is a bit different.  Instead of overriding the specific read/write methods, we implement a virtual file system that is backed by the cloud.  So we deal only in bytes -- text mode etc, appending etc, is dealt with by Pyodide/Emscripten before it talks to us.  

You can look commit by commit or just the whole thing at the end.  Although if you do read one commit at a time, note that I accidentally committed some debug stuff then reverted it shortly after.

It seems to all be working fine with Google Drive; although we can't do automated testing because we don't have an account for that, I've included an (AI-written) SPY file that does some testing of the file system.  I tried testing with One Drive but I think that can't run on localhost?  What's the best way to test that?


PwtKCL (commented):
The code looks fine, I have a very few comment related suggestions, otherwise it's all good to merge.
(There is probably one other change we could make: I had modified Skulpt.js to hook our Cloud FileIO mechanism in Skulpt. This is no longer needed but since Skulpt will eventually be removed from Strype, that can stay for now).

PR #768, opened 2026-03-16
Types cleanup

neilccbrown:
I tried running Typescript's type checking on our project and this MR cleans up most of the errors.  I haven't added the actual command into the CI yet, because there's a few more left that are more serious: the media preview popup needs fixing in Pyodide and causes type errors, plus there's one or two more in the cloud file API work.

But all of these make sense by themselves, I think.  A few small type annotations, and fixes.  The main one is rewriting FrameBody and JointFrames into the new component style.  It turns out that this alone fixes the circular types issue in Typescript, without needing to do Frame.  Since FrameBody and JointFrames are so short, I thought it was worth doing and wasn't too risky in terms of breaking something.  (Bit more hesitant about Frame!)

Oh one more to check: there was a function returning Promise<void> but none of its implementations did anything async, and so it could all safely be turned into void without causing a type error.  But it's possible I misunderstood something.

I'll check when the tests finish that the "expected" ones are failing (hope to sort all that out quite soon!)

This MR is particularly clean to review commit by commit as I was able to split it out nicely.


PwtKCL (approved):
All seems good. 
Regarding the question you asked on the change of type for `onFileToLoadPicked`: you did the right thing. I must have set the returned type as a promise just because it was required somewhere, but it made more sense to do what you did.

PR #772, opened 2026-03-17
Various small fixes

neilccbrown:
This is a set of various small fixes.  It makes most sense one commit at a time.  I wanted to get the tests passing too but I think that will be too big if I keep going.

The first few fixes fix up all the Typescript errors (one at a time), then it adds typechecking to the Github Actions.  I've split out typechecking and linting from Cypress and Playwright, so that if type-checking or linting fails, it still runs the tests alongside.

Then there are a few more small fixes trying to clean up some things that weren't quite working right (Vite and import.meta.glob, the null refs issue I showed you).

Tests will still fail at the moment but hopefully that will be rectified in the next PR or two.


package.json
```diff
@@ -145,10 +146,10 @@
         "start-server-and-test": "^1.15.4",
         "typescript": "^4.9.5",
         "unplugin-vue-components": "^31.0.0",
-        "vite-plugin-checker": "^0.12.0",
+        "vite-plugin-checker": "^0.6.4",
         "vite-plugin-static-copy": "^3.1.6",
         "vue-eslint-parser": "^10.2.0",
-        "vue-template-compiler": "2.7.14"
+        "vue-tsc": "^1.8.27"
```
PwtKCL: Why do we need to remove the package "vue-template-compiler"?

PwtKCL (commented):
I have only a comment update for code change, otherwise some questions.
Mainly, I am worried about removing the package vue-template-compiler.

PR #777, opened 2026-03-20
Pyodide fixes to make tests pass

neilccbrown:
This is nowhere near as big as it looks as many files can be ignored.  I'd look only at the final results because I had a few back and forth.

Things done:
 - Upgrade Github Actions warnings to try to remove deprecation warnings, although we still need a few version of the volta-cli action to fix one more.
 - Swapped to using "Pixi JS Legacy" (which uses Canvas rather than WebGL) as (a) the tests were failing due to WebGL issues and (b) WebGL is generally less available and more finicky, and we don't really need its extra performance just for Turtle rendering.
 - Fixed autocomplete issues around turtle APIs.
 - Increased some test timeouts and slightly refactored a few tests.

I think now all the tests will pass, give or take a couple that are annoyingly flaky due to further timeout issues.


src/stryperuntime/main_bridge_handler.ts
```diff
@@ -158,8 +158,14 @@ export const handleSyncRequests : (
...
-        callbacks.markTurtleDirty();
-        return {request: req.request, response: handleTurtle(turtle, req.buffer).then(() => true)};
+        if (turtle != null) {
+            callbacks.markTurtleDirty();
+            return {request: req.request, response: handleTurtle(turtle, req.buffer).then(() => true)};
+        }
+        else {
+            // PixiJS failed to initialise, so give an error saying that:
+            return {request: req.request, response: Promise.reject("The turtle rendering failed to initialise.  Try refreshing the page or using a different browser.")};
+        }
```
PwtKCL, line 168: Would it be complicated to use a localised version of this error? 
If the error can happen quite often, and localisation is not too complex, I suggest we go for it.
Otherwise we can just ignore it.

PwtKCL (commented):
I have a question regarding some potential localisation. That can be skipped though.
It's fine to merge the code as is otherwise, despite the tests failing, as I suppose test-related fixed are independent, and can do with another PR.

src/stryperuntime/main_bridge_handler.ts
neilccbrown: I think, now we've switched to Pixi JS legacy (which uses Canvas, not WebGL) it shouldn't happen any more.  (I added this when WebGL was failing).  So I think it should be okay; it's no longer an expected error.

PR #778, opened 2026-03-23
Fix service worker

neilccbrown:
Sending this as its own PR as actually I had started a new branch to decouple it from the graphics changes.

As we discussed, this hopefully fixes the service worker issue which seemed to be that sometimes the service worker wasn't taking control of the page.  I now think this may be connected to doing a hard refresh, as the browser prevents service worker connection after this by design (see e.g. https://stackoverflow.com/questions/51597231/register-service-worker-after-hard-refresh ).  This should be a relatively rare problem for us in production (not sure many users hard refresh) but it's nice to have a work-around so we trigger a soft refresh if the page loads after a hard refresh.

We will see if this clears up all the issues with the service worker when we test.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1051,7 +1051,7 @@ export default defineComponent({
                         showMessage: false,
                         readCompressed: true,
                     }
-                ).then(() => {
+                ).then(async () => {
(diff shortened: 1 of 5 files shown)


PwtKCL (approved):
The fix is clear, let's hope it will work fine when we try again.
I merge now, we can see tests later.

PR #779, opened 2026-03-24
Pyodide and Skulpt cleanup

neilccbrown:
This has a few changes:
 - Remove the 808x606 background hack in favour of 800x600.  This all seems to work okay with the current rendering model.
 - Removed special Skulpt related keyboard handling for turtle.  This is the bit I'm least sure about.  I think this may prevent some code movement up and down while the code is running, but I think that's not essential.
 - Made execution switch to the graphics tab if console was showing initially.
 - Reinstated the message about importing turtle/strype.graphics in order to use graphics.  This is painted on the canvas rather than with a message in front, because it messes less with the DOM and seems to work fine.


src/stryperuntime/image_and_collisions.ts
```diff
@@ -74,7 +74,10 @@ export class SpriteManager {
     }
 
     public addSprite(imageOrCanvas : RemoteImage | RemoteCanvas, collidable: boolean, forceId?: number): number {
-        this.dirty = true;
+        // We don't mark dirty for adding the background:
+        if (forceId == undefined || forceId != 0) {
```
PwtKCL, line 78: I think this condition can be simplified by `if (forceId)` as 0 is falsy.

src/components/PythonExecutionArea.vue
```diff
@@ -271,31 +262,9 @@ export default defineComponent({
...
-                    // we hide the graphics container, it will be shown again later when scalling is called
-                    this.graphicsTemporaryHidden = true;
-                    setTimeout(() => {
-                        this.scaleTurtleCanvas(tabContentContainerDiv, graphicsSplitPaneDiv, turtlePlaceholderDiv);
-                    }, 100);
-
-                    // When a canvas has been added we can select the Graphics tab
-                    this.peaDisplayTabIndex = PEATabIndexes.graphics;
-                }
+                this.graphicsImported = (event as CustomEvent).detail as any;
```
PwtKCL, line 265: I think I've seen some time ago we can cast CustomEvent in a neater way: `(event as CustomEvent<boolean>).detail;`

PwtKCL (commented):
The code in PEA seems much clearer now without the Turtle UI part, well done for clarifying all this.
I have 2 minor suggestions over the code, but nothing very important.
(If you want to leave the code as is, and deal the tests fix in another PR, it's fine to merge.)

src/components/PythonExecutionArea.vue
neilccbrown: Ok, this change will be in a future PR.

src/stryperuntime/image_and_collisions.ts
neilccbrown: Not quite; undefined should go into the if (it's "== undefined", not "!= undefined").  I've added a longer comment about what's going on here for the next PR.

PR #780, opened 2026-03-24
Fix the Cloud Dev API key issue

PwtKCL:
This PR is necessary for the Cloud to be used properly as I had made a mistake in handling the Vite modes.
It impacts a lot of the files in the code, but those changes are very specific and clearly identifiable. That's because of the change I made for getting the DEV API keys right: now we are **not** using the "mode" environment variable exposed by Vite to set/get the Strype platform information ("standard" or "microbit").

The reason for this change is because if the mode is set to some specific value, Vite won't look at ".env. development" automatically as the mode won't be detected correctly anymore.
When we don't set the mode, Vite automatically set "development" or "production" for the mode.

So instead, I've used a fully custom environment variable for the Strype platform, so we don't interfere with the Vite mode.
Therefore, the key files to look at are: vite.config.mjs and package.json. Other changes are mostly related to the conditional rendering, replacing "MODE" by "STRYPE_PLATFORM".

**_Important: in your local code, you need to use the API keys in a file called ".env.development" instead of ".env.local" which I mentioned before... that local file was loaded regardless the mode..._**


PwtKCL:
I close this PR: the same content will be added on top of Neil's latest work so I will just recreate a PR.

PR #799, opened 2026-03-27
chore: import type for class that is only being used as type

VividLemon:


VividLemon:
This is only being used as a type, and doesn't need a runtime import

PR #830, opened 2026-04-22
Various bug fixes

neilccbrown:
This is various bug fixes around sound, graphics and console printing.  You can go commit-by-commit but each change is so small that you can probably also just read down the whole thing.


pysrc/python_runner/output.py
```diff
@@ -51,11 +51,13 @@ def should_flush(self) -> bool:
         - It's been at least a second since the last flush
         - The combined length of all the 'text' values is at least 1000 characters
         """
-        return (
-            len(self.parts) > 1
-            or self.last_time and time.time() - self.last_time > self.flush_time
-            or sum(len(p["text"]) for p in self.parts) >= self.flush_length
-        )
+        #return (
```
PwtKCL, line 54: I suppose the comment above can now be removed.
If we always return `True`, do we need the method at all actually?

PwtKCL (commented):
Just a small comment about potentially unused function but all good for merging otherwise.

PwtKCL (approved):

PR #835, opened 2026-04-27
Fixes for graphics library

neilccbrown:
This adds four main things:
 - Key listening for space bar
 - Listening for recent keypress with new get_key() method
 - Fixes get_clicked_actor to work (at all!), and work correctly on overlapping actors
 - Fixes a bug I found during testing where loading from cloud then loading from other places (e.g. demos, local file) would still think it was synced to cloud when we ran Pyodide and it thought there would be a cloud drive.


src/components/PythonExecutionArea.vue
```diff
@@ -536,6 +536,9 @@ export default defineComponent({
             this.appStore.wasLastRuntimeErrorFrameId = undefined;
             // Make sure there is no document selection for our editor
             this.appStore.setSlotTextCursors(undefined, undefined);
+            
+            // Blur run button so space bar doesn't retrigger it:
+            (this.$refs.runButton as HTMLButtonElement).blur();
```
PwtKCL, line 541: It is a sensible idea to remove focus on the run button.
Out of curiosity, once the button loses focus, which part of the UI gets it?

PwtKCL (commented):
When tests pass, the PR is good to merge.
I only have one small question regarding the code behaviour, but otherwise everything is good.

src/components/PythonExecutionArea.vue
neilccbrown: Apparently it goes to the document/<body> element, as a sort of "nowhere in particular" focus.

PR #836, opened 2026-04-27
Handle Google dropouts

### LINKED ISSUE #731: Cloud project: do something about lost of connection when saving a project
When a project is opened long enough (at least with Google Drive) the connection is lost (or reset rather). When a project is then saved, Strype doesn't realise the connection is lost and therefore nothing can be effectively done in term of saving the project.
So when saving a cloud project, we should make sure the connection is still live and do the appropriate thing (reconnection).

PwtKCL:
This PR handles 2 problems that were not handled regarding the use of the Google APIs (fixes #731)

First, we could observe a session dropout pretty soon when no Google related requests were made (~20 minutes). Now having a keep-alive mechanism helps with getting the Google Identity Service session extended longer: until when the access token received from Google expires. When the keep-alive request fails, we try to request a new token.

Second, as mentioned just above, upon authentication to Google, we receive an authentication token that is valid for 1 hour. After that we need to request a new token. This mechanism is therefore intertwined with the keep-alive session.
It seems that Google doesn't provide a way for silently getting this new authentication token: when we need to refresh it, an authentication popup will show (unless the API deems no necessary, but I haven't encounter that).
Note that the popup won't keep showing if the keep-alive interval is reached several time: the Google API knows it's already showing.

### DIFF
--- a/src/components/CloudDriveHandler.vue
+++ b/src/components/CloudDriveHandler.vue
@@ -485,17 +485,18 @@ export default defineComponent({
                         continueSavingProcess();
                     })
                     .catch((responseStatusCode) => {
+                        const errorCode = responseStatusCode.status ?? responseStatusCode;
                         // The following error status codes were relevant for Google Drive. 
                         // We keep them for the general cases, but add a catch up case for other error codes
                         // that may be sent by other Cloud Drives.
                         // Connection issue?
-                        if(responseStatusCode == 401 || responseStatusCode == 403){
+                        if(errorCode){
                             this.proceedFailedConnectionCheckOnSave(cloudTarget);
                             return;
                         }
                         
                         // Folder not found and any other error (400+).
-                        if(responseStatusCode - 400 >= 0){
+                        if(errorCode - 400 >= 0){
                             this.appStore.strypeProjectLocation = undefined;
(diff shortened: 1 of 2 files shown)


src/components/CloudDriveHandler.vue
```diff
@@ -485,17 +485,18 @@ export default defineComponent({
                         continueSavingProcess();
                     })
                     .catch((responseStatusCode) => {
+                        const errorCode = responseStatusCode.status ?? responseStatusCode;
```
neilccbrown, line 491: I found this line a bit confusing initially, as it seemed like an impossible type.  Presumably there's two different places that can throw and one throws the code by itself, the other has it inside a status field?  Might be worth a comment here to explain.

neilccbrown (commented):
One suggestion, otherwise all fine.

src/components/CloudDriveHandler.vue
PwtKCL: This change is likely due to a change of the returned value (when having error) from the Google API. Our code used to expect a number (or a number as string) but when doing tests recently, the type of the response is no longer this.
As this function is inside the generic CloudDriveHandler, I think it's good to keep options as to what return type to expect (for errors).
I'll add a comment to explain this part.

Issue #837, opened 2026-04-28
Flickering frame cursor during operator entry on Chrome


neilccbrown:
This is only in the Pyodide version, and only on Chrome/Chromium.  Make a new project, go to the top of the main code section, press i for an if, and type "a+".  As you type the "+", you will see a frame cursor flicker then disappear.  The same is visible if you add an operator in an assignment statement.  It seems that whenever we add an operator, refactor and refocus, it temporarily focuses the frame cursor then focuses back.  I think this may be behind some of the flakiness in our tests, but even if not, it's quite visible and irritating for users.

PR #842, opened 2026-04-29
Fix missing case where SPY's "blank space" placeholder should be used

### LINKED ISSUE #812: Cannot retrieve a cloud project that has a malformed code
This in an existing issue unrelated to Vue 3/Pyodide changes.

In my code, I have added these lines:
```
a <= "some string"
a.()
```
This is a malformed code, but it can be saved to the Cloud.

However, later if I want to open that project (with that error) from the Cloud, the file won't open because of parsing:
Image

PwtKCL:
This is a single fix PR because we can make an intermediate test version.
It fixes an issue with expressions that contains `<dot operator><blank space><bracket structure>` which were not parsed properly (the blank spaces needed to be translated to the SPY placeholder so that Skulpt can parse that expression without an error.
Fixes #812

### DIFF
--- a/src/helpers/storeMethods.ts
+++ b/src/helpers/storeMethods.ts
@@ -100,11 +100,12 @@ export const generateFlatSlotBases = (slot: { allowedSlotContent?: AllowedSlotCo
             // we have a simple slot, we check if we can infer the detailed type (i.e. number or boolean literals)
             // otherwise we consider it is just a code slot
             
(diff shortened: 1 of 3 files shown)


neilccbrown:
Took me a moment to think about it but yes I think this is the right fix.

Incidentally I think the test failure is caused by bug #837 so I wouldn't worry about it, I will merge in.

PR #843, opened 2026-04-29
Autocomplete fixes

neilccbrown:
The tests finally passed... this fixes all sorts of autocomplete issues.  I'd only look at the final version as I went back and forth a few times fixing some things and breaking other things.  But in the end I think it's cleared up quite a lot of things in the autocomplete.


PwtKCL (approved):
Even if tests are still running, they had passed on Neil's branch as mentioned in the PR description, so I merge it now. All good.

Issue #845, opened 2026-05-01
Duplicate function does not maintain selection


mkolling:
When using 'Duplicate' in the editor, the original selection should be maintained when the content is duplicated (should remain selected), so that you can just hit Ctrl-D multiple times.

(I think this did work initially, but seems not to do it anymore now.)

PwtKCL:
The fix was done, but had not been included in the test version by mistake. It's added now

PR #849, opened 2026-05-11
Various fixes

PwtKCL:
Few fixes around edition + forgotten commit about keeping the frame selection after some menu actions.
(note: these changes with your latest addition for the library support are on the test versions)

### DIFF
--- a/src/components/Frame.vue
+++ b/src/components/Frame.vue
@@ -426,7 +426,7 @@ export default defineComponent({
             // To prevent the command to be called on all frames, but only once (first of a selection), we check that the current frame is a first of a selection.
             // * "this.isPartOfSelection" is necessary because it is only set at the right value in a subsequent call. 
             if(!this.isPythonExecuting && this.isPartOfSelection && (this.appStore.getFrameSelectionPosition(this.frameId) as string).startsWith("first")) {
-                this.duplicate();
+                this.duplicate(true);
             }
         },
 
@@ -441,10 +441,10 @@ export default defineComponent({
                 const anyCanDisable = this.appStore.selectedFrames.some((frameId) => this.canEnableOrDisableFrame(frameId, false));
 
                 if(!anyCanDisable && anyCanEnable){
-                    this.enable();
+                    this.enable(true);
                 }
                 else if(anyCanDisable){
-                    this.disable();
+                    this.disable(true);
                 }
             }
         },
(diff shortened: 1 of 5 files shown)


src/components/Frame.vue
```diff
@@ -525,12 +525,12 @@ export default defineComponent({
...
                 {label: this.$t("contextMenu.downloadAsImg"), onClick: this.downloadAsImg},
-                {label: this.$t("contextMenu.duplicate"), onClick: this.duplicate, attrs: {"action-name": FrameContextMenuActionName.duplicate}, shortcut: isMacOSPlatform() ? "⌘D" : this.$t("shortcut.ctrlPlus") + "D"},
+                {label: this.$t("contextMenu.duplicate"), onClick: () => this.duplicate(false), attrs: {"action-name": FrameContextMenuActionName.duplicate}, shortcut: isMacOSPlatform() ? "⌘D" : this.$t("shortcut.ctrlPlus") + "D"},
                 {divided: "self"},
                 {label: this.$t("contextMenu.pasteAbove"), onClick: this.pasteAbove},
                 {label: this.$t("contextMenu.pasteBelow"), onClick: this.pasteBelow},
                 {divided: "self"},
-                {label: this.$t("contextMenu.disable"), onClick: this.disable},
+                {label: this.$t("contextMenu.disable"), onClick: () => this.disable()},
```
neilccbrown, line 533: Is there a reason for changing the functions into this style?  For the `duplicate` function I understand a param was needed, but for disable and enable does it make a difference?

src/components/Frame.vue
```diff
@@ -1142,13 +1142,14 @@ export default defineComponent({
...
 
-        duplicate(): void {
+        duplicate(keepSelection?: boolean): void {
             if(this.isPartOfSelection){
                 this.appStore.copySelectedFramesToPosition(
                     {
                         newParentId: (this.isJointFrame)
                             ? getParentId(this.appStore.frameObjects[this.frameId])
                             : getParentOrJointParent(this.frameId),
+                        keepSelection: keepSelection,
```
neilccbrown, line 1152: It doesn't matter much, but I think you can write `keepSelection,` and it will automatically put the variable in with that name?  Or does that not work if the parameter is optional.

neilccbrown (approved):
All fine, just a couple of small stylistic notes, but not necessary to address them, just notes for you to think about or to understand if I've missed something..

src/components/Frame.vue
PwtKCL: Good point. I had the habit of always prefixing the properties with their name, I didn't realise TS supports this shorthand notation. I'll amend it for a future PR.

src/components/Frame.vue
PwtKCL: Yes, that's because TS expects onClick to be of this type:` (e?: MouseEvent | KeyboardEvent) => void` as defined by the context menu library.

PwtKCL:
Note: the last commit fixes #845

PR #850, opened 2026-05-11
Match case statements support

### LINKED ISSUE #826: Add support for Python's "match"
We should be able to support the Python's `match` statement (i.e. frame wise, parsing wise).
The new frame should come with one blank `case`.
Make sure that deleting `case` if that's the only one triggers an error (a `match` must contain at least one `case`).

PwtKCL:
This PR contains changes to support the "match-case" statements from Python 3.10 (fixes #826).

For Strype it means: having new frames, adding new error check rules (for match frames: they have to contain at least one valid "case" frame), prevent the deletion of match and case frames using backspace (from their body).
Also, as we handle the "empty" match error as a precompiled error, we need to discard that same error from TigerPython.

For Skulpt: the current version of Skulpt doesn't support the parsing of match/case frames.
I've updated the Skulpt library for that, with the minimal changes required for the parser.
The parser is a bit permissive to allow empty match to technically be saved and loaded in Strype, as well as comments and blanks in match. As we may get rid of Skulpt at some point, I don't think there will be a fork + CR for Skulpt?
Note that I've also removed everything about the old FileIO mechanism, since the Skulpt library was recompiled from the source code.

Finally, the frame colours have changed as discussed recently to keep the same colour for similar "meta-type" of frames.


src/helpers/pythonToFrames.ts
```diff
@@ -1255,6 +1266,38 @@ function copyFramesFromPython(p: ParsedConcreteTree, s : CopyState) : CopyState
...
+        // This case not supported by original Skulpt version - so to limit the changes in Skulpt, the Skulpt parser
+        // is permissive and allows cases (normal) + pass (for us) + simple statements (for us).
+        // The simple statements are therefore limiting the accepted content to things like "a", "a()", "a=b", but not if or while etc.
+        // So we make a check in checkValidMatchContent() when parsing the children of a match statement to cover the permissive Skupt version.
+        const r = makeAndAddFrameWithBody(p, AllFrameTypesIdentifier.match, 0, [1], 3, s);
+        s = r.s;
+        break;
+    }
+    case Sk.ParseTables.sym.case_block: {
+        console.log("getting a block ?");
```
neilccbrown: Leftover debug statement?

neilccbrown (commented):
Generally looks fine.

I think we might have to swap to loading Python from within Python sooner rather than later, but it is going to be a bit of a pain to do.

src/helpers/pythonToFrames.ts
PwtKCL: oops, yes

PwtKCL:
As Neil mentioned, the Playwright tests failing are most likely due to #837. For Cypress, it's not clear why the test fails but it doesn't seem to be related to the source code changes, so I'll merge and we can investigate deeper if it fails continuously.

Issue #315, opened 2024-10-30
Add autocompletion for image and sound file names


neilccbrown:
With the new addition of image and sound files in the graphics world, we need some way to find them.  I think as a first step, rather than an image library we could have autocomplete for them in string literals (like Greenfoot does for project files), showing the image in the right-hand side.

PR #854, opened 2026-05-12
Autocompletion for built-in file names in string literals

neilccbrown:
We now have files available in Strype in the /strype directory, but there's no way to see a list of them.  This adds that ability via the autocomplete mechanism, as already suggested in #315.  Looks like this when you hit ctrl-space in a string literal:

image

### DIFF
--- a/src/components/AutoCompletion.vue
+++ b/src/components/AutoCompletion.vue
@@ -92,6 +92,21 @@ import microbitDescriptions from "@/autocompletion/microbit.json";
 import microbitAPI from "@/autocompletion/microbit-api.json";
 // #v-endif
 
+const assetFileList : string[] = Object.keys(import.meta.glob(
+    "/src/assetsFilesystem/**/*",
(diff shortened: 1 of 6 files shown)


src/components/AutoCompletion.vue
```diff
@@ -310,7 +325,11 @@ export default defineComponent({
             this.showFunctionBrackets = true;
             const imported = await getAllExplicitlyImportedItems(context);
             this.acResults = {};
-            if (token === null) {
+            if (token != null && kind == "string") {
+                this.acResults["Files"] = assetFileCompletions;
```
PwtKCL: I suppose we'll want this module to be localised.
Should it be "Built-in files" to be even more precise?
Here are the two possible translations for French: "Fichiers" or "Fichiers intégrés".

PwtKCL (commented):
Apart a suggestion for localisation, all good as it is.

Issue #844, opened 2026-05-01
Add another keyboard input function: get_key()


mkolling:
I think I'd like another get_key() function: one that waits until a key is pressed and returns it (returns the _next_ key, not the last).

If we do this, I think we should name this "get_key()" and rename the other one "get_last_key()".

PR #855, opened 2026-05-13
Change get_key to block and wait for next key

neilccbrown:
As discussed in the meeting, removes the "old" (we never released it) get_key and replaces it with a version that blocks to wait for the next key.  I've done it in two commits: a removal and addition, in case we later decide we want both and want to rollback the removal.  But it's probably simplest to just read the overall diff to see the change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18385,23 +18385,6 @@
             "license": "ISC",
             "peer": true
         },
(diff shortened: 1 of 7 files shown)


PwtKCL (approved):

PwtKCL:
note: this addresses #844

Issue #645, opened 2025-11-11
Add coordinate system info about Image for the Strype Media API


PwtKCL:
The coordinates specified in Image's functions are relative to the top left corner of the image.
I think this applies for all the functions of Image.

We should explain this in the doc, either at the Image class descriptor or at every function that uses the image's coordinate system.

This is important because the "world" coordinate system works differently.

_Note that we may want **to first work on a data science project**, and then decide if we leave this issue for a specific DS API as the Media API might just be used for gaming purpose then._

PR #857, opened 2026-05-13
Add support for matplotlib and update some documentation

neilccbrown:
It turns out that matplotlib support was quite easy, following along based on WebTigerPython.  You can see what it looks like in the screenshots that the tests use.

One particular note: I deliberately suppressed the "Import turtle or strype.graphics for graphics" message when matplotlib is loaded but didn't add mention of matplotlib.  I think we want that message to be targeted at novices and first-time users, and matplotlib is too advanced for them so we don't want to suggest it.

Also note that matplotlib takes a long time to load, just because it has so many dependencies and Pyodide takes a while to load them all.  Hence printing a progress message.  We can always think about an alternate progress display for library loading in future, or other ways to do this.  (We could actually pre-load these libraries for next execution ahead of time, and change when the imports change, but it's a lot of work for something that only more advanced users will be doing.)


PwtKCL (approved):
Excellent !! For the record, this PR also handles the issue #645 and #853.
Let's see when tests are finished to merge the PR, just in case.

PwtKCL:
The tests are not passing - some of them are related to another thing so for the moment, let's merge this PR so we can focus on the tests side separately.

PR #863, opened 2026-05-18
Version update

PwtKCL:
Version-related changes in preparation of v2.1.0


PwtKCL:
As this is only related to versioning I'm merging this PR.

PR #864, opened 2026-05-18
Fix multiple issues

neilccbrown:
This fixes several bugs:
 - say_for was broken due to setTimeout not working in the web worker with Pyodide
 - Syntax errors were not printing a message  (and weren't processing the message to highlight a line) and we weren't printing stderr in Pyodide
 - Add a function to clean console in a new strype.builtins module that is auto-imported
 - Extended the clone method on Image to support rotation and flipping


src/stryperuntime/image_and_collisions.ts
```diff
@@ -82,7 +84,7 @@ export class SpriteManager {
         }
         const id = forceId ?? this.nextSpriteId++;
         const box = collidable ? this.collisionSystem.createBox({x:0, y:0}, imageOrCanvas.width, imageOrCanvas.height, {isCentered: true}) : null;
-        const newImage = {id, img: imageOrCanvas, x: 0, y: 0, rotation: 0, scale: 1, collisionBox : box};
+        const newImage = {id, img: imageOrCanvas, x: 0, y: 0, rotation: 0, scale: 1, collisionBox : box, removeAtTime: null};
```
PwtKCL, line 87: I'm a bit surprised TS accepted the type of `removeAtTime`. Based on the next lines, should it be `null | number` ?

src/stryperuntime/image_and_collisions.ts
```diff
@@ -93,27 +95,44 @@ export class SpriteManager {
...
 
     public hasSprite(id: number) : boolean {
-        return this.sprites.has(id);
+        const sprite = this.sprites.get(id);
+        if (sprite !== undefined && (sprite?.removeAtTime == null || sprite.removeAtTime < Date.now())) {
+            return true;
+        }
+        else {
+            // Remove it if it was present, as it was scheduled for it:
+            this.sprites.delete(id);
```
PwtKCL, line 104: Just to be sure: would this be silently work even if the ID isn't contained in the list of sprites?

pysrc/strype/graphics.py
```diff
@@ -398,23 +398,33 @@ def draw_polygon(self, points):
...
             copy = Image(self.get_width(), self.get_height())
             copy.draw_image(self, 0, 0)
         elif scale <= 0:
             raise ValueError("Clone scale must be greater than zero")
+        elif flip is not None and flip != "horizontal" and flip != "vertical":
+            raise ValueError("Clone flip must be \"horizontal\", \"vertical\", or None")
         else:
-            copy = Image(self.get_width() * scale, self.get_height() * scale)
-            copy._draw_part_of_image(self, 0, 0, 0, 0, self.get_width(), self.get_height(), scale)
+            copy = Image(-42, -42)
```
PwtKCL, line 425: What are the literal values here?

PwtKCL (commented):
I just have one main observation about the literal values, otherwise it's only comments about TS or code understanding.

src/stryperuntime/image_and_collisions.ts
neilccbrown: I think maybe it knows that null can be generalised into something with null | number ?

src/stryperuntime/image_and_collisions.ts
neilccbrown: Yes, that's right.

pysrc/strype/graphics.py
neilccbrown: They are magic numbers used to indicate we'll set the image ourselves.

PwtKCL:
As discussed over Teams yesterday, the failing tests are the usual timeouts related problems so we can merge.

PR #872, opened 2026-05-26
Got clear_console showing in code completion.

neilccbrown:
A small PR to fix a bug with clear_console not being in the code completion (it's in an automatically imported module strype.builtins so we needed a bit of extra work so that the code completion knows about this.)

Fixes #869


PwtKCL (approved):
I think the tests will be fine with the change, so I merge now.

PR #874, opened 2026-05-26
Fix issue with Firefox not saving on reload/close

neilccbrown:
While starting to look at changing our browser storage model I came across a case where the current implementation was failing on Firefox.  It seemed twofold: Firefox wasn't triggering "beforeunload", and it wasn't finishing the async action to save the state.  This matches with issues I've seen sometimes on Firefox where it didn't seem to remember changes I made shortly before reloading.

I've changed which event we listen to and stopped saving being async by removing the checksum.  It would be possible to alternatively fix this by using a Javascript package for checksumming (which is sync, not async), but I'm not sure we really need the checksum?  We set a checksum then stringify then compress.  If we manage to uncompress successfully and find out the checksum is altered, I'm not sure what that means or what we should do about it.  So I favour removing it but happy to put it back if we have a reason to want it.


PwtKCL:
Yes it's a good idea to remove the checksum. 
The change is fine, it might need small adjustments for the tests to pass, so it's OK to merge once they are added

PR #875, opened 2026-05-29
Change storage model for Strype

neilccbrown:
I want to do a bit more testing after this (and add more tests) and add in a way to reload old discarded projects.  But I'm worried the PR will grow too large so I'm submitting this partway stage.  The tests seem a bit flaky today but I think it's just flakiness and no longer any "real" failures.

Anyway: as we discussed recently, this changes Strype's storage model.  When you open a new tab, it opens a fresh project.  Each tab holds its own state, both in memory and saved regularly as before, although this time to IndexedDB rather than localStorage.  The reason is that we now store multiple simultaneous projects and so it's more likely we'd hit the 5-10MB limit on localStorage, especially once image literals are involved.  This does make for one awkward corner case.

The awkward case is that when you close the browser tab, we want to save the current state.  But writing to IndexedDB is async, and it's very likely (observed for real on Firefox) that the browser kills the pending async part before it gets a chance to write.  So for this one case we store the "emergency-saved" state in localStorage and later on a page load we move such emergencies to IndexedDB.

The async restriction also led to changing how new projects work because we don't want to wait for an IndexedDB read while loading the app.  So we always load the empty state, then load the "new" state [print(myString) etc] from App.mounted().  I don't see a flicker on Firefox from doing this.

If you refresh the browser tab it retrieves the same project, by having a tab ID that lives in sessionStorage which outlives the refresh.  We no longer need certain dialogs, for example when you open a share link we don't need to ask about replacing the existing content because that doesn't apply any more.  This change in particular has caused the diff to appear larger than it really is, because it meant de-indenting a bunch of code in App.vue that used to be in callbacks but it hasn't actually changed.  You will get a better view if you hide whitespace changes while reading (handy link: https://github.com/k-pet-group/Strype/pull/875/changes?w=1 )

I also tidied up the Cypress tests a bit by moving duplicated beforeEach code into a common helper, as I changed how it waits for page load (instead of waiting for <body> to load, it actually waits for 2+ frames to appear.  This makes sure Cypress doesn't start pressing keys in the few milliseconds between <body> loading and the "new" project state being inserted; this had caused flaky tests when I initially changed things.

(I'd review this only by looking at the final changes, as I went back and forth a bit on a few things.)


src/store/store-db-storage.ts
```diff
@@ -0,0 +1,196 @@
...
+
+            if (!db.objectStoreNames.contains(STORE)) {
+                db.createObjectStore(STORE, {
+                    keyPath: DatabaseFieldNames.tabId,
+                });
+            }
+        };
+
+        request.onsuccess = () => resolve(request.result);
+        request.onerror = () => reject(request.error);
```
PwtKCL, line 46: Here and in the other promises of the page, how is reject actually handled by us? Should we have a indication showing to the user when we fail to do any of the operations?

src/store/store.ts
```diff
@@ -20,35 +20,19 @@ import $ from "jquery";
...
-            isExistingStateLocated = true;
-            returnedState = initialStates["initialEmptyState"];        
-        }
+export function getEditorTabId() : string {
+   
+    let tabId = sessionStorage.getItem(AutoSaveKeyNames.strypeEditorTabId);
+
+    if (!tabId) {
+        // None found, generate one and save it:
+        tabId = crypto.randomUUID();
```
PwtKCL, line 29: That seems to be the best way to get an ID. Out of curiosity, is there any API browsers implements that would expose a tab ID as assigned by the browser itself?

src/store/store.ts
```diff
@@ -20,35 +20,19 @@ import $ from "jquery";
...
-        // #v-else
-        returnedState = initialStates["initialMicrobitState"];
-        // #v-endif
-    }
-    return (returnedState as StateAppObject);
+    return tabId;
 }
 
-const initialState = getState();
+const initialState = initialStates["initialEmptyState"] as StateAppObject;
```
PwtKCL, line 35: So in a way, we are doing very similar thing as before (to avoid the flickering) since we load an empty step first: I think then that's why no flicker showed when you tried.

PwtKCL (commented):
Just a few comments but otherwise it's fine.
I would also suggest to check the localised literals no longer required after dropping the modal dialog.
However, I'm not entirely sure about discarding that dialog (even if we said it should be). In the end, isn't there a risk user lose their code if they open a shared project? With younger users, don't we want to make sure they won't accidently lose their work?

src/store/store.ts
neilccbrown: I asked the AI, but it suggested using sessionStorage to implement it; apparently there's not a built-in API to do it.

src/store/store-db-storage.ts
neilccbrown: This is a good point.  I think what I may do is a second PR to add error handling.

neilccbrown:
> However, I'm not entirely sure about discarding that dialog (even if we said it should be). In the end, isn't there a risk user lose their code if they open a shared project? With younger users, don't we want to make sure they won't accidently lose their work?

When they open a sharing link, it's usually in a new context, because either you opened it in a new tab, or you clicked on a link from a website that wasn't the Strype editor.  So in those cases there's no prior code to lose because any existing code is in another tab and won't be overwritten.  The case where code can be lost is if you close a tab without saving, or load a new project from an existing tab, and the dialogs will remain for those cases.

PwtKCL:
> > However, I'm not entirely sure about discarding that dialog (even if we said it should be). In the end, isn't there a risk user lose their code if they open a shared project? With younger users, don't we want to make sure they won't accidently lose their work?
> 
> When they open a sharing link, it's usually in a new context, because either you opened it in a new tab, or you clicked on a link from a website that wasn't the Strype editor. So in those cases there's no prior code to lose because any existing code is in another tab and won't be overwritten. The case where code can be lost is if you close a tab without saving, or load a new project from an existing tab, and the dialogs will remain for those cases.

I see, that makes sense. I am so used to copy links during dev from a work-in-progress on Strype that I didn't think of this more common workflow...

PR #877, opened 2026-06-01
Small fixes

PwtKCL:
This PR only contains 2 basic fixes that were found out during the preparation work of the "rename variable" new feature.
Fixes #876, we could not paste a global statement (as Python text) inside a function definition frame.

### DIFF
--- a/src/helpers/pythonToFrames.ts
+++ b/src/helpers/pythonToFrames.ts
@@ -508,7 +508,7 @@ function transformCommentsAndBlanks(codeLines: string[], format: "py" | "spy") :
 // ready to be pasted immediately afterwards.
 // If successful, returns a map with key-value Strype directives.  If unsuccessful, returns a string with some info about
 // where the Python parse failed.
-export function copyFramesFromParsedPython(codeLines: string[], currentStrypeLocation: STRYPE_LOCATION, format: "py" | "spy", linenoMapping?: Record<number, number>, dryrun?: "dryrun" | undefined, ignoreGlobalRuleOut?: boolean) : string | null | Map<string, string> {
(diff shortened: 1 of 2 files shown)


src/helpers/pythonToFrames.ts
```diff
@@ -1559,7 +1564,9 @@ export function pasteMixedPython(completeSource: string, clearExisting: boolean)
         err = copyFramesFromParsedPython(s.defs, STRYPE_LOCATION.DEFS_SECTION, s.format, s.defsMapping, "dryrun");
     }
     if (typeof err != "string") {
-        err = copyFramesFromParsedPython(s.main, STRYPE_LOCATION.MAIN_CODE_SECTION, s.format, s.mainMapping, "dryrun");
+        // We have a small ambiguous situation to handle: pasting global statements should not raise an error if we are actually inside a function definition
+        const ignoreGlobalRuleOut = curLocation == STRYPE_LOCATION.IN_FUNCDEF;
+        err = copyFramesFromParsedPython(s.main, STRYPE_LOCATION.MAIN_CODE_SECTION, s.format, s.mainMapping, "dryrun", ignoreGlobalRuleOut);
```
neilccbrown: I wonder if instead of passing the global flag, we should instead pass curLocation as the code location (2nd param in the call), instead of MAIN_CODE_SECTION?  I think that would stop us banning global, and avoids adding the next parameter.  But I may have missed something that would stop this working.  Might be worth a quick try?

neilccbrown (commented):
I've suggested an alternate approach that might avoid having to have a new parameter, but happy to discuss if there's a reason that wouldn't work.

src/helpers/pythonToFrames.ts
PwtKCL: As discussed yesterday, the change is a good idea and should be fine. I've pushed a new commit, the manual tests I've made worked, let's see how the CI tests pass to make sure.

PR #879, opened 2026-06-02
Added error handling for when indexedDB is unavailable, and added a test for it

neilccbrown:
As discussed earlier, add error handling for indexedDB issues along with an error message.  The Playwright test confirms that it does show.

I thought I'd separate this off as a small PR rather than lump it in with the next bit of work about loading old orphaned states.


PwtKCL (commented):
It seems fine, except the small forgotten imports of main.ts that made the linter tests fail. Once this is fixed, I think it's fine to merge even if the other tests may still fail.

PR #880, opened 2026-06-03
User analytics caputuring

Htzzz11:
Implementation of user analytics capture. Refer to [Analytics_summary.md] for a comprehensive summary of all the analytics implemented and the batching mechanism.


src/components/OpenDemoDlg.vue
```diff
@@ -164,10 +164,14 @@ export default defineComponent({
...
 
-        getSelectedDemo() : ({ name : string, demoFile: Promise<string | undefined> } | undefined) {
+        getSelectedDemo() : ({ name : string, demoFile: Promise<string | undefined>, source: "builtin" | "mediacomp-strype" } | undefined) {
             if (this.selectedDemoItemIndex >= 0 && this.selectedDemoItemIndex < this.demosInCurrentCategory.length) {
                 const d = this.demosInCurrentCategory[this.selectedDemoItemIndex];
-                return {name: d.name, demoFile: d.demoFile()};
+                const selectedDemoGroup = this.availableDemos[this.selectedDemoCategoryIndex];
+                const source = (selectedDemoGroup?.type === "Library" && selectedDemoGroup?.name === "mediacomp-strype")
+                    ? "mediacomp-strype"
+                    : "builtin";
```
neilccbrown, line 173: It is possible for us to add more external demos in future (and the user can add them themselves now if they add a new library address in the demos dialog).  So the code here should be adjusted to:

    const source = (selectedDemoGroup?.type === "Library" && selectedDemoGroup?.name)
                    ? selectedDemoGroup?.name
                    : "builtin";

Then adjust the return type of the function to be "source: string"

src/helpers/analyticsCountry.ts
```diff
@@ -0,0 +1,71 @@
...
+            : null;
+        return {countryCode, countryName};
+    }
+    return UNKNOWN_COUNTRY;
+}
+
+export async function fetchUserCountry(): Promise<UserCountry> {
+    for (const lookupUrl of COUNTRY_LOOKUP_URLS) {
+        try {
+            const response = await fetch(lookupUrl, {
```
neilccbrown: I'd suggest switching to use axios.get here (we already have a dependency on axios), which will allow you to set a timeout for this fetch (maybe 3 seconds?).  As it stands I think this fetch is not time-limited so it can wait a while in the worst case.

src/helpers/bootstrapApp.ts
```diff
@@ -0,0 +1,36 @@
+import type { Pinia } from "pinia";
+import { settingsStore, useStore } from "@/store/store";
+import { startSessionTracking } from "@/helpers/sessionTracker";
+import { Analytics_batch_flush_ms } from "@/helpers/analyticsConstants";
+
+export async function bootstrapApp(pinia: Pinia): Promise<void> {
```
neilccbrown: I think this name is unfortunate in Strype as we also use the Bootstrap library for some GUI stuff.  I'd call the function (and file) "initialiseAnalytics" instead.

src/helpers/bootstrapApp.ts
```diff
@@ -0,0 +1,36 @@
...
+    const store = useStore(pinia);
```
neilccbrown: You don't actually need to pass a pinia param to useStore or settingsStore, you can just call `useStore()` or `settingsStore()` once they're initialised.  So you can drop the "pinia" parameter to this function.

src/helpers/bootstrapApp.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (import.meta.env.DEV) {
+        (window as unknown as {useStore: () => typeof store}).useStore = () => store;
```
neilccbrown: I don't know if this relates to the pinia parameter confusion, but I suspect this is also unnecessary, as everywhere can just call useStore() anyway.

src/helpers/sessionTracker.ts
```diff
@@ -0,0 +1,46 @@
+import type { useStore } from "@/store/store";
+import { Analytics_session_idle_threshold_ms, Analytics_session_tick_ms } from "@/helpers/analyticsConstants";
+
+export function startSessionTracking(store: ReturnType<typeof useStore>): void {
```
neilccbrown: You can drop the store param here, and just call useStore().

src/helpers/sessionTracker.ts
```diff
@@ -0,0 +1,46 @@
...
+        sessionEnded = true;
+        tick();
+        store.analyticsFrameCount = Object.values(store.frameObjects).filter((f) => f.id > 0).length;
+        store.enqueueAnalyticsEvent("session_end", {
+            activeDurationMs: store.analyticsActiveSessionTime,
+            frameCount: store.analyticsFrameCount,
+        });
+    };
+
+    window.addEventListener("beforeunload", emitSessionEnd);
```
neilccbrown: In bootstrapApp you have handlers for these events which flush the queue, and here you have handlers for these events that add to the queue.  I wouldn't trust the ordering to do the right thing; I'd combine these handlers and the handlers from bootstrapApp so that they are a single handler which runs once (beforeunload and pagehide can both run, so your instinct here to have a boolean flag to guard against multiple runs is correct) and does the right order: enqueue a session_end then flush.

src/store/store.ts
```diff
@@ -218,6 +229,30 @@ export const useStore = defineStore("app", {
...
+
+            analyticsPlatform: "editor" as "editor" | "microbit",
+
+            analyticsLocale: "" as string,
+
+            analyticsEventQueue: [] as AnalyticsEvent[],
+
+            analyticsPendingOutputChars: 0 as number,
+
+            analyticsFlushInProgress: false as boolean,
```
neilccbrown: This is my fault for not thinking about this ahead of time, but I think this is probably best outside the store now I see the full code.  This can be a object somewhere, possibly still declared in this file, at the top.  Then it will be a matter of find/replace in this file on "this.analytics" to e.g. "analytics.analytics".  I think that may be all that's needed, so not too bad to change?  You can keep the analytics methods in the store for this PR.

src/store/store.ts
```diff
@@ -709,6 +744,222 @@ export const useStore = defineStore("app", {
...
+                return;
+            }
+            this.enqueueAnalyticsEvent("locale_change", {from: previousLocale, to: newLocale});
+            this.analyticsLocale = newLocale;
+        },
+
+        async initAnalyticsCountry() {
+            const country = await fetchUserCountry();
+            this.setAnalyticsCountry(country);
+            console.log("Country analytics:", {
```
neilccbrown: In general, we try not to log to the console unless it's critical info; this can be removed.

src/store/store.ts
```diff
@@ -1291,7 +1542,9 @@ export const useStore = defineStore("app", {
         updateState(newState: Record<string, unknown>){
             //this method complete changes the state with a new state object
             Object.keys(this.$state).forEach((property) => {
-                (this as any)[property] = newState[property];
+                if (property in newState) {
+                    (this as any)[property] = newState[property];
+                }
```
neilccbrown: This change should be unneeded with moving analytics out of the store, so revert this bit.

src/store/store.ts
```diff
@@ -2401,7 +2654,16 @@ export const useStore = defineStore("app", {
...
-            
+
+            // Strip analytics fields — these are session-scoped and re-initialised by bootstrapApp on every load.
+            // Persisting them would cause sessionId, active-session timers, throttle timestamps, and per-session
+            // counters to leak across reloads.
+            Object.keys(stateCopy).forEach((key) => {
+                if (key.startsWith("analytics")) {
+                    delete stateCopy[key];
+                }
+            });
```
neilccbrown: Ditto: This change should be unneeded with moving analytics out of the store, so revert this bit.

(It's partly for this reason that we will move them out of the store.)

src/types/vue-component-api-types.ts
```diff
@@ -98,7 +98,7 @@ export type CaretContainerComponentAPI = {
 }
 
 export type OpenDemoDlgComponentAPI = {
-  getSelectedDemo: () => ({ name : string, demoFile: Promise<string | undefined> } | undefined),
+  getSelectedDemo: () => ({ name : string, demoFile: Promise<string | undefined>, source: "builtin" | "mediacomp-strype" } | undefined),
```
neilccbrown: To match earlier comment, the source type here will become string.

src/main.ts
```diff
@@ -81,8 +82,9 @@ const app = createApp(App);
 app.use(i18n);
 
 // Store package (Pinia is the default store management library for Vue 3)
-app.use(createPinia());
-
+const pinia = createPinia();
+app.use(pinia);
+void bootstrapApp(pinia);
```
neilccbrown: As per earlier comment, you won't need to store the pinia here.

src/helpers/sessionTracker.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    document.addEventListener("mousemove", onActivity, { passive: true });
+    document.addEventListener("keydown", onActivity, { passive: true });
+    document.addEventListener("click", onActivity, { passive: true });
+    document.addEventListener("scroll", onActivity, { passive: true });
+
+    const tick = () => {
+        const now = Date.now();
+        if (now - lastActivityTime < Analytics_session_idle_threshold_ms) {
+            store.analyticsActiveSessionTime += now - lastTickTime;
```
neilccbrown: I think this logic can overcount if tick is suspended.  Browsers may suspend setInterval calls in inactive tabs, so there is a possibility that this happens:

    last tick before browser suspends sets lastTickTime
    <a day passes>
    user returns and immediately does something, this executes and adds a day to the active session time

I think if you make the right-hand side `min(now - lastTickTime, Analytics_session_tick_ms)` it will avoid this issue?

src/helpers/bootstrapApp.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+    store.initAnalyticsUserId();
+    store.initAnalyticsSession();
+    store.initAnalyticsPlatform();
+    startSessionTracking(store);
+    await store.initAnalyticsCountry();
```
neilccbrown: I think we shouldn't await here.  We do then risk the country code not being available immediately, but we want to complete the rest of the initialisation just below before waiting for the country.  We could put it at the end and await, but that's no real difference to calling it here and not awaiting.

neilccbrown (commented):
This is good work.  There are a few things to fix, and one organisational thing to adjust that I should have decided earlier: in retrospect I think it's right that the analytics items live outside the store in just a standard top-level object in the store file.  I think with a find/replace it won't be so hard to adjust this.  But you will find it allows you to remove some of your added logic in the store.

src/store/store.ts
Htzzz11: Agreed, thanks. We moved it out. All the analytics state and logic is now in src/store/analytics.ts module. Certainly is cleaner, also allow us to drop the guard in updateState and the strip analytics block in getStateJSONStr.

src/helpers/sessionTracker.ts
Htzzz11: Good catch, tick now caps the increment at one interval with `Math.min(now` - lastTickTime, `Analytics_session_tick_ms),` so a suspended background tab can't dump a large gap into the active session time on the tick that fires when it wakes up.

src/helpers/bootstrapApp.ts
Htzzz11: Agreed, I think we overlooked this since it was one of our earliest changes. Removed await as suggest.

src/helpers/sessionTracker.ts
Htzzz11: Combined into a single guarded handler: `onAnalyticsPageUnload()` in sessionTracker.ts runs once (module-level boolean guard) and does the final tick → captures frame count → enqueues session_end → `flushAnalyticsQueue("unload")` in that order. `startSessionTracking()` no longer registers its own listeners; `initialiseAnalytics` registers this one handler on beforeunload, pagehide, and visibilitychange→hidden.

PR #881, opened 2026-06-05
Add a banner to offer to reload recently unsaved state

neilccbrown:
If you close a window with state that was changed after its last save, and you open a new editor window soon afterwards (note: I increased this timeout a lot for dev mode, so you can reload recent state more easily), a banner is offered as to whether you want to load that state or not.  If you choose either way, the state is marked so that it's not offered again.

If there are multiple recent states (unlikely but can happen) only the most recent is offered, and all older are marked so they won't be offered.

The plan is still to offer all the recent such states in the file open menu but this PR grew big enough I think it's worth submitting now.  A couple more items:
 - I also updated Playwright as it seems they slightly improved the tracing in the more recent versions.
 - I improved service worker loading in Playwright in Firefox, which may help a few flaky tests around that issue.


src/components/MessageBanner.vue
```diff
@@ -77,6 +85,20 @@ export default defineComponent({
             this.appStore.currentMessage = MessageDefinitions.NoMessage;
         },
 
+        onKeyDown(e: KeyboardEvent) : void {
+            // Only look for F1, F2 etc:
+
+            const match = e.code.match(/^F(\d{1,2})$/);
```
PwtKCL, line 91: While this is a good idea for accessibility (more that development I think), is it guaranteed to work in Strype? 
Some browsers may "force capture" the functions keys and not let us interact with them.
(Some issues may happen for debugger keys but i doubt we'll reach more than 3 buttons in a message anyway, so that's OK)

src/store/store.ts
```diff
@@ -177,7 +179,9 @@ export const useStore = defineStore("app", {
 
             projectName: i18n.global.t("defaultProjName"),
 
-            isEditorContentModified: false,
+            isEditorContentModified: false, // Whether the content is modified since last external (disk/cloud/share) save
```
PwtKCL, line 182: While this is used principally for the cases mentioned in the comment, we still set the flag regardless the origin of the project.

PwtKCL (commented):
Just 2 small remarks but all good otherwise - I've reloaded the failed tests of piece of mind, but the PR can be merged beforehand if it's fine without them to complete.

PR #884, opened 2026-06-07
Allow loading of old unsaved states from the Open dialog

neilccbrown:
At the bottom of the Open dialog, we now show a list of old states that can be loaded.  We can discuss/tweak the exact rules in future but this will allow testing and discussion.

Once this is merged I suggest we put up the test server so that we can test the saving changes and the analytics.


src/components/Menu.vue
```diff
@@ -736,7 +736,7 @@ export default defineComponent({
             this.recentLoadableStates = (await checkForRecentSaveStates(settingsStore().locale ?? "en", "load_menu"))
                 .map((s) => {
                     return {
-                        label: `Recent state (${ceil(s.data.length / 1024)} KB)`,
+                        label: `${s.projectName} (${ceil(s.data.length / 1024)} KB)`,
```
PwtKCL, line 739: "KB" can also be localised. Typically, in French for example, it would show "Ko".

PwtKCL (commented):
Good to merge. 
I have a small comment about localisation.
You can address it later, I'll keep the PR open for now so you can address the comment, either making the change now or later (and then you can merge).

PwtKCL:
(As suggested by Neil, I'll make the translations in a separate PR - so this one is now merged.)

PR #896, opened 2026-06-10
Add a dialog for the book, and book projects

neilccbrown:
As mentioned in the meeting, this adds a new dialog (similar to the open demos dialog, but different enough it wasn't worth trying to share the same code) for the book, which lets you access the book projects and some media files too.

It's not as big as it looks; you can ignore everything in the public directory, and just look at the actual Typescript code changes.

Other changes:
 - Fixed that loading bug I showed you, which turned out to be whenever the frame cursor was in the imports or definitions when triggering the load of a book (or demo) project.
 - Changed the demo and book dialog to only ask about discarding the current project when you actually choose the new project to load; if you cancel either dialog it keeps the current project state rather than discarding before opening the dialog.  This is necessary for the book dialog especially, as you might be going just to copy a sound/image file and you don't want to lose your code in that case.


PwtKCL (approved):
I'll add the French translation later in my next PR.

PR #901, opened 2026-06-11
Asset filesystem changes for v2.1.1

neilccbrown:
As discussed, this is a set of changes against v2.1.0 which just changes the file paths on the assets and makes them live at e.g. /books, /sounds rather than /strype/sounds etc.  Removing the prefix is a little bit more work as we have to mount separate "filesystems" for each top-level dir, so one for /books, one for /sounds, which are parameterised by their mountpoint ("books", "sounds", etc).  The plan is to release this as 2.1.1.

The plan is to merge this into main as well after it has been reviewed.


PwtKCL (approved):
As confirmed by Neil, the failing Cypress test is fixed in the current main branch. So the PR can be merged as is.

Issue #838, opened 2026-04-28
Find and replace


PwtKCL:
Having a find & replace functionality in Strype could be a good thing.

PR #903, opened 2026-06-11
Rename identifiers feature

PwtKCL:
This PR contains an implementation for a "rename identifier" feature. 
It is a popup (Bootstrap popover) showing when an identifier name change is detected.
The detection and display are based on several factors:
- changes are made in specific label slots structures (in specific frame types),
- the number of identifiers before/after in the slots structure are the same (this is to avoid some intermediate editing actions triggering the popup on the one hand, and proposing irrelevant changes on the other hand)
- there are potential changes in the editor (based on name matching and a few rules depending on the context)

The popup action can be triggered with the keyboard (ctrl/⌘+r) or via click on the button.
It can be discarded when a click outside the popup is detected, or when the escape key is pressed, when the frame cursor moves or when text is typed.

image

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -569,8 +569,12 @@ export default defineComponent({
                 return;
             }
 
-            // The "rename" identifier tool can be executed when we are editing or not. But never when Python is running. 
-            if (!this.isPythonExecuting && (event.ctrlKey || event.metaKey) && event.key ==  "r"){
+            // The rename identifier action cannot be executed when Python is executed or code is edited.
+            // (These actions should already discard the popup, but just to make sure...)
+            // We only want to capture the rename identifier action keyboard shortcut if one popup is visible,
+            // otherwise we leave the shortcut for the browser.
+            const openedPopup = document.querySelector(".popover.show:has(." + scssVars.renameIdentifierPopoverClassName + "");
+            if (openedPopup && !this.isPythonExecuting && !this.appStore.isEditing && (event.ctrlKey || event.metaKey) && event.key ==  "r"){
                 eventBus.emit(CustomEventTypes.renameIdentifier);
                 event.preventDefault();
                 event.stopPropagation();
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,7 +1,7 @@
 <template>
     <!-- With the new package for Bootstrap (for Vue 3), BApp must wrap the application content -->
     <BApp>
-        <div>
+        <div @mousedown.capture="onAppMouseDownCapture">
             <div v-if="showAppProgress || setAppNotOnTop" :class="{'app-overlay-pane': true, 'app-overlay-pane-absolute': showAppProgress, 'app-progress-pane': showAppProgress}" @contextmenu="handleOverlayRightClick">
                 <div v-if="showAppProgress" class="app-progress-container">
                     <div class="progress">
@@ -117,7 +117,7 @@
 //////////////////////
 //      Imports     //
 //////////////////////
-import { defineComponent } from "vue";
+import { defineComponent, watch } from "vue";
 import { useI18n } from "vue-i18n";
 import { BApp } from "bootstrap-vue-next";
 import MessageBanner from "@/components/MessageBanner.vue";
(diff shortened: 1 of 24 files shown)


src/components/LabelSlotsStructure.vue
```diff
@@ -154,6 +176,14 @@ export default defineComponent({
...
         });
         eventBus.on(CustomEventTypes.updateParamPrompts, this.updateParamPromptsIfInList);
+        eventBus.on(CustomEventTypes.renameIdentifier, this.renameIdentifiers);
+        eventBus.on(CustomEventTypes.closeOtherRenameIdentifierPopup, this.closeRenameIdentifierPopupIfNotActive);
+    },
+
+    unmounted(){
+        // Just to be safe with events, we clear off any registrations
+        eventBus.off(CustomEventTypes.updateParamPrompts, this.updateParamPromptsIfInList);
+        eventBus.off(CustomEventTypes.renameIdentifier, this.renameIdentifiers);
```
neilccbrown, line 186: Should we also clear the listener for closeOtherRenameIdentifierPopup?

src/components/LabelSlotsStructure.vue
```diff
@@ -1129,6 +1187,125 @@ export default defineComponent({
...
         },
+
+        async checkAndShowRenameIdentifiersPopup() {
+            // We may show the popup when :
+            // 1) we are on a label slots struct that reflects the corresponding frame's label type (FrameLabel) "isIdentifierDefsContainer" property is set to true,
+            // 2) we have the same number of elements in both old and new identifiers list
+            // 3) we detect that some of the variables' (old) name is used in the code
+            const thisFrame = this.appStore.frameObjects[this.frameId];
+            const frameType = thisFrame?.frameType;
+            if(thisFrame && frameType && frameType.labels[this.labelIndex].isIdentifierDefsContainer){
```
neilccbrown: One of the CI tests failed here with this exception:

```
  > Cannot read properties of undefined (reading 'isIdentifierDefsContainer')

When Cypress detects uncaught errors originating from your application it will automatically fail the current test.

This behavior is configurable, and you can choose to turn this off by listening to the `uncaught:exception` event.

https://on.cypress.io/uncaught-exception-from-application
      at Proxy.checkAndShowRenameIdentifiersPopup (http://localhost:8081/editor/src/components/LabelSlotsStructure.vue:1109:68)
      at Proxy.blurEditableSlot (http://localhost:8081/editor/src/components/LabelSlotsStructure.vue:1045:9)
      at _createElementBlock.onBlur._cache.<computed>._cache.<computed> (http://localhost:8081/editor/src/components/LabelSlotsStructure.vue:1242:81)
      at callWithErrorHandling (http://localhost:8081/editor/node_modules/.vite/deps/vue.js?v=a4e84b16:1231:17)
      at callWithAsyncErrorHandling (http://localhost:8081/editor/node_modules/.vite/deps/vue.js?v=a4e84b16:1238:15)
      at HTMLDivElement.invoker (http://localhost:8081/editor/node_modules/.vite/deps/vue.js?v=a4e84b16:6950:3)
      at remove (http://localhost:8081/editor/node_modules/.vite/deps/vue.js?v=a4e84b16:11458:23)
      at performRemove (http://localhost:8081/editor/node_modules/.vite/deps/vue.js?v=a4e84b16:5354:4)
      at remove (http://localhost:8081/editor/node_modules/.vite/deps/vue.js?v=a4e84b16:5362:10)
      at unmount (http://localhost:8081/editor/node_modules/.vite/deps/vue.js?v=a4e84b16:5330:18)
```

I don't immediately see how, but it is during unmount (and during a reload); I wonder if the store gets cleared then we get called for loss of focus?  But it's odd the frames still exist but the labelIndex doesn't.  Might be worth a precautionary `this.labelIndex < frameType.labels.length` check.

src/helpers/storeMethods.ts
```diff
@@ -155,6 +156,80 @@ export const retrieveSlotByPredicate = (frameLabelSlotStructs: SlotsStructure[],
...
+    }
+    else{
+        const foundSlots: SlotCoreInfos[] = [];
+
+        // Look in the frame, its child and joint frames
+        const allFramesToLookIn = [lookAndRenameInFrameId, ...getAllChildrenAndJointFramesIds(lookAndRenameInFrameId)];
+        allFramesToLookIn.forEach((frameId) => {
+            // If a frame is disabled or is frozen (or part of a frozen ancestor), we ignore the change.
+            const thisFrame = useStore().frameObjects[frameId];
+            if(thisFrame.isDisabled || useStore().isEffectivelyFrozen(frameId)){
```
neilccbrown: Disabled I agree not to change, but frozen... I wonder if we should.  If it's really using those variables/names I don't see that the user would want to not update them as it's just bound to cause an error afterwards.

src/App.vue
```diff
@@ -544,6 +568,15 @@ export default defineComponent({
                 event.stopPropagation();
                 return;
             }
+
+            // The "rename" identifier tool can be executed when we are editing or not. But never when Python is running. 
+            if (!this.isPythonExecuting && (event.ctrlKey || event.metaKey) && event.key ==  "r"){
```
neilccbrown: I'm a bit confused here why this doesn't change if a rename popup is showing?  As it stands isn't it forever capturing ctrl-r and blocking it, even if there's no popup showing?

tests/cypress/e2e/load-save.cy.ts
```diff
@@ -106,14 +98,14 @@ function testEntryDisableAndSave(commands: string|string[], disableFrames: strin
     });
 } 
 
-describe("Loads and re-saves fixture files", () => {
+describe.only("Loads and re-saves fixture files", () => {
```
neilccbrown: Accidental .only commit; need to remove this

tests/cypress/e2e/load-save.cy.ts
```diff
@@ -106,14 +98,14 @@ function testEntryDisableAndSave(commands: string|string[], disableFrames: strin
...
     it("Loads a basic project", () => {
         testRoundTripImportAndDownload("tests/cypress/fixtures/project-basic.spy");
     });
     it("Loads a basic trisection project", () => {
         testRoundTripImportAndDownload("tests/cypress/fixtures/project-basic-trisection.spy");
     });
-    it("Outputs a dummy for solo try", () => {
+    it.only("Outputs a dummy for solo try", () => {
```
neilccbrown: Ditto

neilccbrown (commented):
I've reviewed the code.  You'll have seen that the Playwright tests failed, and I think it's worth trying to get them to pass.  In particular, MacOS failed several of the rename tests on all the browsers.  I tried running locally on my Mac and saw the same.  At first I thought it was the Ctrl/Meta+R thing but actually it's the arrow keys.  If you change all the "Control+ArrowRight" to "ControlOrMeta+ArrowRight" I think they'll pass.  (You can also change the renameKBShortcut thing to "ControlOrMeta+r" if you like; it's a built-in fake key name in Playwright exactly for this Windows vs Mac issue.)

I notice that in Ubuntu+Webkit we have a lot of flaky tests (100+) but that is not new to your branch, that also occurs in main.  Whereas Ubuntu+Chromium has 0 flaky tests, Mac+Webkit has only 10.  I might take another look at why the failure rate is so high.

src/components/LabelSlotsStructure.vue
PwtKCL: Yes, it's always better to keep them clean, thanks for pointing it out.

src/helpers/storeMethods.ts
PwtKCL: Good point. My initial rationale for that was that users would make changes that they cannot see (or at least not just by navigating in the editor).
But you're right: better that than generating errors.

tests/cypress/e2e/load-save.cy.ts
PwtKCL: Sorry and thanks for pointing it out again. I'll add a mechanism to prevent ".only" in Cypress to run tests successfully in the CI. We have this for Playwright in place, so it's good to find a similar way in Cypress.
For the sake of checking, I'll remove the ".only" of this file in a second commit, so we can check the mechanism works as expected.

src/App.vue
PwtKCL: I've added a control on the popup visibility to only capture the keyboard shortcut when one popup is showing.

PwtKCL:
This fixes #838

PR #905, opened 2026-06-15
Fix various issues in the new storage model

neilccbrown:
Fixes a few issues with the work to change over a storage model:
 - We now (again!) show the "Are you sure you want to leave this page?" browser dialog if they try to leave while the content is unsaved.
 - Added a "Clear all" link in the open dialog that gets rid of all old unsaved states.
 - Only show the header in that dialog if there are actually unsaved states.
 - Fixed a bug where reloading more than once would lose the state (because after first reload, it's considered an old unmodified state so we were clearing it out before looking for states to load).
 - Stopped showing the "Load recent state?" banner when you do New project (presumably, you want a new project!)
 - Changed which handler we do the emergency save in after some issues in Firefox.


tests/playwright/e2e/storage-model.spec.ts
```diff
@@ -401,6 +401,60 @@ test.describe("Offer to reload unsaved backups", () => {
             await loadAndWaitForEditor(page6);
             await assertOpenRecentMenu(page6, []);
         });
+        
+        test(`Load two states, optionally save 2nd (${state2Saved}) then use new project, should be no banner`, async ({browser, browserName}) => {
+            const context = await browser.newContext({recordVideo: {dir: "tests/playwright/test-results/videos/"}});
+            const page1 = await context.newPage();
+            console.log("Page1 video: " + await page1.video()?.path());
+            page1.on("console", (msg) => console.log("Browser log page 1:", msg.text()));
```
PwtKCL, line 409: I suppose these two lines (and similar for page2 can be deleted?)

PwtKCL (commented):
Just a comment regarding logs in a test - otherwise I think it can be merged.

tests/playwright/e2e/storage-model.spec.ts
neilccbrown: Ah those are actually wanted in case of any future CI failures.  The bottom line is needed in Playwright to see any console output from the page being tested, and the top line is needed if you want to see the path to the video; Playwright records separate videos per page in multi-page tests but doesn't label them so it's very hard to debug unless we label them like this.

PR #912, opened 2026-06-18
Fix various scrolling issues

neilccbrown:
Fixes several scrolling issues:
 - Runtime errors should be scrolled into view, into the centre if scrolling was needed
 - Ditto for the focused item after undo/redo
 - Console should scroll to show the new output if it was already at the bottom

Also fixed a minor thing that "strype.builtins" should not show in the autocomplete for import/from-import, because it's automatically imported already.


tests/playwright/e2e/scroll-into-view.spec.ts
```diff
@@ -42,6 +45,57 @@ async function expectWellInsideViewport(locator: Locator, margin = 100) {
...
+    while ((match = regex.exec(input)) !== null) {
+        const text = input.slice(lastIndex, match.index);
+        const key = match[1];
+
+        if (text) {
+            await page.keyboard.type(text);
+        }
+
+        await page.keyboard.press(key);
+
```
PwtKCL, line 93: I suppose both the type and press method may need to be followed by a timeout, as keyboard input is notoriously not working well without in Firefox

tests/playwright/e2e/scroll-into-view.spec.ts
```diff
@@ -42,6 +45,57 @@ async function expectWellInsideViewport(locator: Locator, margin = 100) {
...
+        lastIndex = match.index + match[0].length;
+    }
+
+    // remaining text after last token
+    const remaining = input.slice(lastIndex);
+    if (remaining) {
+        await page.keyboard.type(remaining);
```
PwtKCL, line 100: Same here

src/components/PythonExecutionArea.vue
```diff
@@ -648,7 +648,14 @@ export default defineComponent({
                         pythonConsole.value = "";
                     }
                     else {
-                        pythonConsole.value = pythonConsole.value + output;
+                        // We want to append, then only scroll if they were already near the bottom:
```
PwtKCL, line 651: Why do we want to check the scrolling position? Don't we want to always scroll down when some output is showing inside the console?

PwtKCL (commented):
A suggestion for some test function that might be helping Firefox tests (if they fail), and question regarding the console scrolling.
Otherwise it seems good.

src/components/PythonExecutionArea.vue
neilccbrown: Essentially, if the user has scrolled up manually (which is the only way it will not be at the bottom, with the new system) we don't want to interrupt that.  If they want to continue from the bottom they can scroll back down.  Without this if you're an infinite loop you can't scroll or stop it scrolling for you, which can be frustrating.

PR #924, opened 2026-06-26
Use text clipboard for frame copy/paste

### LINKED ISSUE #753: Feature Request: Exporting single frames (f. e. classes) instead of the whole project
Currently, Strype allows exporting the entire project as a .py file. I was wondering if it could be made possible to export and import single frames (f. e. classes) individually.

In a project-like setting I want my students to work on different parts of a larger project (e.g., one student develops the "Athlete" class, another the "Club" class). If we could export/import single frames, the code could easily merged into one main project without having to manually clean up or overwrite the entire codebase.  

Image

neilccbrown:
Previously we used an internal clipboard for copy/paste, which prevented copying and pasting between Strype windows.  This change makes copy/paste use the text clipboard, by using our existing mechanism for pasting Python code, and exporting to Python.  The frame wrapping and dragging (which used the old copy/paste) have also been adjusted.

It's a bit big but there wasn't really a good halfway point.  Oh and I would definitely only look at the end, I had to make a few adjustments along the way so individual commits go back and forth.

Fixes a related feature request; we don't need a single frame export now they can be copied to clipboard.

Fixes #753


src/components/CaretContainer.vue
```diff
@@ -246,6 +239,22 @@ export default defineComponent({
             // so we check that 1) we are on the caret position that is currently selected and 2) that paste is allowed here.
```
PwtKCL: I suppose the comments here (above and below) need adjustment, since we no longer "pre"-check the validity of the frames to be pasted at the location in the same way than before - and that we don't wipe the clipboard after copying frame anymore

src/components/CaretContainer.vue
```diff
@@ -246,6 +239,22 @@ export default defineComponent({
...
             // We should know the action is about pasting frames or text if some text is present in the clipboard (we clear it when copying frames)
             if (this.isFocusedForPaste) {
+                let pasteDestination = {id: this.frameId, caretPosition: this.caretAssignedPosition};
+                // If we currently have a selection of frames, the pasted frame should replace the selection, so we delete that selection.
+                // (it should be fine regarding the grammar check because the caret will be at the same level whether it's before or after the selection)
+                if(this.appStore.selectedFrames.length > 0){
```
PwtKCL, line 242: I'm not clear about this: so yes, the caret is at the same level, but how does it mean the grammar is guaranteed to be correct?

src/components/CaretContainer.vue
```diff
@@ -176,7 +170,6 @@ export default defineComponent({
         return {
             CustomEventTypes, // just for using in template
             scssVars, // just to be able to use in template
-            showPasteMenuItem: false,
             insertFrameMenuItems: [] as StrypeContextMenuItem[],
```
PwtKCL, line 173: I forgot about what we said in meeting regarding the menu entry being available when there is nothing to paste. Should we have it only shown when text or media we support are on the clipboard? It's most likely the case but if nothing is to be pasted, it feels wrong to have the menu (or show it as something can be tentatively pasted).
That's a bit of a headache of course because there is no reactivity on the clipboard as such, but I guess with a method it's possible to achieve it, if we want to do it. 
Another option: show a "nothing to paste" message?

src/helpers/editor.ts
```diff
@@ -1102,14 +1103,8 @@ const bodyMouseUpEventHandlerForFrameDnD = (event: MouseEvent): void => {
...
             if(event.ctrlKey || event.altKey){
-                if(currentDraggedSingleFrameId){
-                    useStore().doCopyFrame(currentDraggedSingleFrameId);
-                    useStore().pasteFrame({clickedFrameId: currentCaretDropPosFrameId, caretPosition: currentCaretDropPosCaretPos});
-                }
-                else{
-                    useStore().doCopySelection();
-                    useStore().pasteSelection({clickedFrameId: currentCaretDropPosFrameId, caretPosition: currentCaretDropPosCaretPos});
-                }
+                const text = copyFrameTextReadyForClipboard(currentDraggedSingleFrameId ? [currentDraggedSingleFrameId] : useStore().selectedFrames);
```
PwtKCL, line 1106: While the logic is totally making sense with the new code, I wonder if in this very specific case using the store manipulation isn't a better choice in term of efficiency? I'm not sure, it would imply working on a object copy and traverse the tree to change IDs so maybe it's still a lot. Besides, it avoid any problem with pasteMixedPython if there were some, to impact this functionality.

src/helpers/editor.ts
```diff
@@ -2347,3 +2342,36 @@ export function isFullyInViewport(el: Element, margin = 0) : boolean {
...
+    if (childrenIds.length > 0) {
+        return {id: childrenIds[childrenIds.length - 1], caretPosition: CaretPosition.below};
+    }
+    else {
+        return {id: parentId, caretPosition: CaretPosition.body};
+    }
+}
+
+// frameIds must be contiguous
+export function copyFrameTextReadyForClipboard(frameIds: number[]) : string {
```
PwtKCL, line 2358: Very tiny remark, I think the comment can say that frameIds must _represent_ contiguous frames; otherwise it reads as the numbers should be contiguous.

src/helpers/editor.ts
```diff
@@ -2347,3 +2342,36 @@ export function isFullyInViewport(el: Element, margin = 0) : boolean {
...
+        code = p.parse({
+            startAtFrameId: frameIds[0],
+            stopAt: {frameId: frameIds[frameIds.length - 1], includeThisFrame: true},
+        });
+    }
+    return code;
+}
+
+// Must be called in response to click or key handler so that we have access to the clipboard
+// frameIds must be contiguous
```
PwtKCL: Same comment as before.

src/helpers/editor.ts
```diff
@@ -2347,3 +2342,36 @@ export function isFullyInViewport(el: Element, margin = 0) : boolean {
...
+export function copyFramesToClipboard(frameIds: number[]) : void {
+    let code = copyFrameTextReadyForClipboard(frameIds);
+    navigator.clipboard.writeText(code).catch((err) => {
+        console.error(err);
```
PwtKCL, line 2375: We might want to make a clear message here. The odds are probably low that writing to clipboard fails, but if it does, it is quite a disruptive problem. So I think it can be shown as an error message in Strype proper.

src/helpers/pythonToFrames.ts
```diff
@@ -1420,11 +1463,11 @@ function makeMapping(section: NumberedLine[]) : Record<number, number> {
     }, {} as Record<number, number>);
 }
 
-// Takes a list of lines of Python code and splits them into three sections: imports, function definitions, and main code.
+// Takes a list of lines of Python code and splits them into four sections: imports, function definitions, class definitions, and main code.
 // Each line of the original will end up in exactly one of the three parts of the return.
```
PwtKCL: Small adjustment to this line: "one of the _four_ parts of the return."

src/helpers/pythonToFrames.ts
```diff
@@ -1504,47 +1553,76 @@ export function splitLinesToSections(allLines : string[]) : {projectDoc: string[
...
+        classDefsMapping : makeMapping(classDefs),
         mainMapping : makeMapping(main),
         headers: {} as Record<string, string>,
         format: "py",
     };
 }
 
 // Returns headers if successful, or null if there was an error (which will already have been shown in the UI)
-export function pasteMixedPython(completeSource: string, clearExisting: boolean) : { headers: Record<string, string> } | null {
+export function pasteMixedPython(completeSource: string, at: CurrentFrame, clearExisting: boolean = false, dontSetCaretAfter = false) : { headers: Record<string, string> } | null {
```
PwtKCL: Tiny small detail, either remove ":boolean" from the 3rd parameter or add it to the forth for consistency.

src/store/store.ts
PwtKCL, line 70: This isn't related to this work, but I've been wondering if we should get rid of property, and only use the method instead as we do already to get the next ID? It seems safer and more consistent. I don't think we really need to have this property anywhere in the code as such since we can compute it when we need it.

src/parser/parser.ts
```diff
@@ -562,6 +562,8 @@ export default class Parser {
         return this.parse({startAtFrameId: useStore().getImportsFrameContainerId, stopAt: {frameId: useStore().getDefsFrameContainerId, includeThisFrame: false}});
     }
 
+    // You can only pass start if you also pass stop.  (But stop by itself is fine)
```
PwtKCL, line 565: Why is start without stop not allowed?

src/store/store.ts
```diff
@@ -538,36 +531,13 @@ export const useStore = defineStore("app", {
...
-                        return true;
-                    }
-                }
-    
-                return false;
-            };            
-        },
-        
         // frameToBeMovedId is an optional argument and it is used in cases where we are just checking if a 
         // frame can be moved to a position based on the copied frame type --> we are not really checking about the actual copied Frame
```
PwtKCL: This comment is no longer valid.

src/store/store.ts
```diff
@@ -538,36 +531,13 @@ export const useStore = defineStore("app", {
...
         isPositionAllowsFrame() {
-            return (targetFrameId: number, targetCaretPosition: CaretPosition, lookingForTargetPos: boolean, frameToBeMovedId?: number) => {
+            return (targetFrameId: number, targetCaretPosition: CaretPosition, lookingForTargetPos: boolean, frameToBeMovedId: number) => {
                 // Where do we get the frame from --> from copiedFrames if it is a copied frame
                 // Otherwise the input frame is to be checked (e.g. for moving an else statement or duplicating an else statement -- which doesn't go anywhere).
```
PwtKCL: The comments here can also be adjusted.

src/store/store.ts
```diff
@@ -2019,18 +1927,15 @@ export const useStore = defineStore("app", {
 
             if (this.selectedFrames.length > 0 && frame.allowChildren) {
                 this.isWrappingFrame = true;
-                this.copySelection();
+                const text = copyFrameTextReadyForClipboard(this.selectedFrames);
                 // For deleting a selection, we don't care if we simulate "delete" or "backspace" as they behave the same
```
PwtKCL, line 1927: Same thing that for Drag & Drop, I'm not sure which is the best in terms of performance and code safety.

src/store/store.ts
```diff
@@ -2730,281 +2635,63 @@ export const useStore = defineStore("app", {
...
-            cloneFrameAndChildren(sourceFrameList, payload.frameId, payload.newParentId, {id: nextAvailableId}, copiedFrames); 
-
-
-            // Add the copied objects to the FrameObjects
-            Object.keys(copiedFrames).map(Number).forEach((id: number)=> {
-                this.frameObjects[id] = copiedFrames[id];
-            });
             
-            const topFrame = copiedFrames[Object.keys(copiedFrames).map(Number)[0]];
+            const stateBeforeChanges = cloneDeep(this.$state);
```
PwtKCL, line 2640: There are certainly many places in the code where we could have save a clone if we don't need it. Maybe it is still good to generate stateBeforeChanges if we are going to use it, as the state can be large.  It would of course require a bit of checking in the counterpart place where we use it, but it might be better in the end.

src/store/store.ts
```diff
@@ -2730,281 +2635,63 @@ export const useStore = defineStore("app", {
...
-            // Add the top frame to the its new parents children list
-            childrenListToBeAdded.splice(
-                payload.newIndex,
-                0,
-                topFrame.id
-            );
+            const areFramesJoint = payload.sourceFrames.frames[payload.sourceFrames.frameIds[0]].frameType.isJointFrame;
+            if (areFramesJoint) {
+                // Joint frames, so we know we're inserting below.  We need to see if the frame we have is
+                // the joint parent (the "if") or a joint child (e.g. "elif"):
```
PwtKCL: That's certainly what you pointed out on Teams: joint frames are pasted from inside a root/joint body last position

src/types/types.ts
```diff
@@ -326,7 +326,7 @@ export interface EditableFocusPayload extends SlotCoreInfos {
 export interface NavigationPosition {
     frameId: number;
     isSlotNavigationPosition: boolean, // flag to indicate if we are working with a slot position (change from previous version that used composite types)
-    caretPosition?: string;
+    caretPosition?: CaretPosition;
```
PwtKCL, line 329: After so many years ahhaah! Sorry :)

PwtKCL (changes requested):
I think the changes look good in general, thanks for working on this great improvement (both for the UI and the code base... I think some of the weird frame paste methods are because of me...).

I mainly have comments about old comments that can be adjusted, and some question about what should be done for Drag & Drop and wrapping. It makes sense we use the same logics, but I don't know if that is the best choice ultimately.

There are 2 bugs I've identified from trying the branch:
- we can paste an elif frame joint to a while loop frame (I think this isn't a new bug, it could happen before with text copy already) - but now we can also paste one after if/else (and we lose the frame caret)
- trying to paste elif inside import: doesn't do anything (good) but doesn't show a message (should it?)
-trying to paste elif inside main does nothing (good) but lose the frame caret (and again should we show a message?)
- starting from the default project, copy a single word (text copy from outside Strype) and paste it inside the main section with the keyboard: it works fine. Now do the same with trying to paste via the context menu: it doesn't, no paste occurs. 
- copy varassign frames, go to defs and paste: they go in main

Here is the example for an elif frame:
```
elif True :
	print("b")
```

src/components/CaretContainer.vue
neilccbrown: Ok, will fix.

src/components/CaretContainer.vue
neilccbrown: This comment came from the code that got inlined from an existing function... so I think that's not my comment.  But I think what it's saying is after the selection is deleted it doesn't matter whether the cursor was before or after it?

src/components/CaretContainer.vue
neilccbrown: I think in general it's not well supported to pre-emptively check the clipboard for valid content.  There's also a few potential issues: (a) the content shown when the menu is posted could change if they alt-tab during, and (b) the content could be e.g. a megabyte long string that takes a long time to decide if it's valid Python.  So we decided we will just allow the paste menu option and see what happens we attempt the paste.

src/helpers/editor.ts
neilccbrown: I did briefly wonder about this too, but decided in the end that it's more maintainable to just have the one pathway for copy and paste and re-use it.  We can revisit it if we notice problems.

src/helpers/editor.ts
neilccbrown: Good point, fixed.

src/helpers/pythonToFrames.ts
neilccbrown: Fixed

src/helpers/pythonToFrames.ts
neilccbrown: Fixed

src/parser/parser.ts
neilccbrown: I've added a comment to explain this.

src/store/store.ts
neilccbrown: Removed

src/store/store.ts
neilccbrown: Fixed that now.

src/types/types.ts
neilccbrown: Haha, I just looked at it and thought I think that could be a CaretPosition.  The type checked didn't complain so I think it's fine...

neilccbrown:
Not urgent but just for info: I've added some new commits to solve the flagged issues.  In particular:
 - I've fixed a bug you mentioned (which was in the tracker already) that pasting assignments into definitions went into main.
 - I've added a system to decided if a paste is valid for joint frames.  This is an idea from the AI, to use regexes as a simple way to write the rules (e.g. ifs are: /if,elif*,else?/).  So when you are considering a paste, you glue the frame types together and see if it matches the regex.  It seems to work fine.

src/components/CaretContainer.vue
PwtKCL: Oh ok sorry I didn't notice it was there before. 
Since before we were checking for the possibility of pasting beforehand, that comment was to say "it doesn't matter about being before or after the selection when we want to check the grammar since both positions are on the same level in their parent". 
But now that might be useless to specify it as we don't check rules the same way. I think it can be removed.

src/components/CaretContainer.vue
PwtKCL: Yes, I was thinking just to make a basic check as "there is text" or "media content", but not in details. But the odds the clipboard is empty (or of different content nature) are probably low compared to when there is something that could be pasted (as action - even if it doesn't result in frame pasted as such).
For the reasons you mentioned anyway let's forget about it.

src/types/types.ts
```diff
@@ -636,7 +645,7 @@ export function generateAllFrameDefinitionTypes(regenerateExistingFrames?: boole
             { label: "if ", defaultText: i18n.global.t("frame.defaultText.condition") },
             { label: " :", showSlots: false, defaultText: "" },
         ],
-        allowJointChildren: true,
+        allowJointChildren: `${j(StandardFrameTypesIdentifiers.elif)}*${j(StandardFrameTypesIdentifiers.else)}?`,
```
PwtKCL, line 648: I wonder if we should try to generate the string values programmatically based on the property `jointFrameTypes`. If we could, it would avoid any redundant code writing and avoid mistakes. 
We have a method called `generateAllFrameCommandsDefs` that's called at least once when Strype is loaded that already set some parts of the definitions for us (based on locale for that case, but it can be extended to more generic things).

PwtKCL (commented):
I have one suggestion about generating `allowJointChildren`, and noticed a small bug that as mentioned by Teams (pasting to empty joint root's frames).
Otherwise I think once that bug is fixed it will be good to merge.
The other pending comments are generic stuff so they're not required for the PR.

src/store/store.ts
neilccbrown: I think that's sensible but won't include it in this PR as it's already big.

PwtKCL (approved):
All good for merging

PR #931, opened 2026-06-30
Actor API changes

neilccbrown:
This is a set of Actor API changes from last week's meeting:
 - Add a `remove_actors` function taking an optional tag
 - Add an `Actor.re_add` method which takes an X and Y position
 - Allow `set_location` to take None for X or Y, to leave them untouched
 - Some tweaks to the "touching" methods, most easily seen in the code diffs

Also added a couple of tests for the new bits.


PwtKCL (commented):
I think the test which failed isn't related to the changes so I merge now.

PR #934, opened 2026-07-02
Assorted small fixes

neilccbrown:
This will make most sense by commit, I think.  It has various small fixes, some of which were bugs I noticed while trying to get matplotlib interactivity working, as I was pasting in example code from their website.

Features added:
 - Added the (x, y) in the PEA header while hovering over the graphics canvas while not executing.
 - Added mouse support to matplotlib, it now responds to mouse move and click events (may need more improvement in future; we don't currently have useful APIs for mouse drags or listening to all keys without blocking, maybe we want new public or private APIs for this)

Parsing/pasting bugs/features:
 - Support for * as unary operator, this is for parameter expanding, e.g. "f(*a)"
 - Fixed trailing commas being discarded, Python uses these for single item tuples, e.g. "(a,)"
 - Don't allow blanks to be pasted in a class def since you're not allowed to insert them there
 - Allow "as" in from import, e.g. "from a import b as c"

### DIFF
--- a/src/workers/python-execution.ts
+++ b/src/workers/python-execution.ts
@@ -277,6 +277,9 @@ if ${usingMatplotlib ? "True" : "False"}:
             def show(self):
                 _figure_renderer_pyodide.render_current_figure(self.canvas.figure)
                 
+            @classmethod
+            def start_main_loop():
+                
                 # Now run a main loop:
                 x, y = -1, -1
                 import strype.graphics as g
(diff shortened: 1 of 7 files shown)


src/workers/python-execution.ts
```diff
@@ -382,7 +435,9 @@ runner`);
                 // So no need to scale.
                 
                 // Then add it as a sprite (default 0, 0 position is fine):
+                console.log("Plot size: " + image.width + " x " + image.height);
```
PwtKCL: Maybe removed if not intended to be left for information

PwtKCL (commented):
Just a comment about a console.log.
Otherwise, let's wait for adding the "**" operator for rest support as we discussed ealier and see how tests behave, then it will be good to merge.

PR #938, opened 2026-07-03
Fix "global" parser

PwtKCL:
We have overlooked that global can take more than 1 variable:  `global myString, yourString` is a valid statement in Python.
Consequently, there were 2 issues in Strype when the above situation occurred: we didn't parse the Skulpt tree correctly for "global" statements, and we didn't parse the editor's content to Python properly.
This PR fixes #937


src/helpers/pythonToFrames.ts
```diff
@@ -1119,6 +1122,8 @@ function copyFramesFromPython(p: ParsedConcreteTree, s : CopyState) : CopyState
     case Sk.ParseTables.sym.return_stmt:
         // Return may or may not have an expression child after it:
         if (children(p).length >= 2) {
+            console.log("for return: ");
```
neilccbrown: Leftover debug output?

tests/cypress/fixtures/project-global-multi-vars-1.spy
```diff
@@ -0,0 +1,13 @@
+#(=> Strype:1:std
+#(=> Section:Imports
+#(=> Section:Definitions
+def testF (toAdd1,toAdd2 ) :
+    global myString,yourString
```
neilccbrown, line 5: Might be worth also having a global with one arg and a global with say four args, just to check those are also fine.

neilccbrown (commented):
Couple of small suggestions but after that if the tests pass, fine to merge.

src/helpers/pythonToFrames.ts
PwtKCL: Yes sorry

PR #939, opened 2026-07-06
Cloud pyodide improvements

PwtKCL:
This PR addresses two things:
- a fix to handle the Cloud files opened with "append" mode via Pyodide (#928)
- a new helper Strype builtins function, `get_connected_cloud`, that reaturns the name of the Cloud provider the Strype project lives in (or None if not relevant).

The docs and tests have also been updated (included some tests for generic Pyodide file IO operations - which are not completely relevant for Strype as such but might still be handy to have).

### DIFF
--- a/src/stryperuntime/main_bridge_handler.ts
+++ b/src/stryperuntime/main_bridge_handler.ts
@@ -195,6 +195,8 @@ export const handleSyncRequests : (
     }
     case "getCurrentCloudName":
         let cloudName = undefined;
+        // The following values are directly returned to the user code, so we should NOT modify them in the future
+        // to prevent any incompatibility within the older user programs which may have used them...
         switch (useStore().syncTarget) {
         case StrypeSyncTarget.gd:
             cloudName = "Google Drive";
--- a/src/stryperuntime/main_bridge_handler.ts
+++ b/src/stryperuntime/main_bridge_handler.ts
@@ -12,6 +12,7 @@ import {useStore} from "@/store/store";
 import { handleTurtle, TurtlePixiHandler } from "@/stryperuntime/turtle_pixi_handler";
 import { sInput } from "@/helpers/execPythonCode";
 import {getRawFileFromLibraries} from "@/helpers/libraryManager";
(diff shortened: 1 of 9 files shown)


src/stryperuntime/main_bridge_handler.ts
```diff
@@ -192,6 +193,19 @@ export const handleSyncRequests : (
             return {request: req.request, response: Promise.reject("The turtle rendering failed to initialise.  Try refreshing the page or using a different browser.")};
         }
     }
+    case "getCurrentCloudName":
+        let cloudName = undefined;
+        switch (useStore().syncTarget) {
+        case StrypeSyncTarget.gd:
```
neilccbrown, line 201: Maybe just add a comment here saying these values are directly returned to user code, so we shouldn't modify them.  (To prevent future accidents!)

src/stryperuntime/pyodide-emscripten-cloud-fs.ts
```diff
@@ -98,6 +98,11 @@ export function getFSForEmscripten(pyodide: PyodideAPI) : EmscriptenFileSystemPl
             return content.length;
         },
         write(stream: FSStream, buffer: Uint8Array | Int8Array, offset: number, length: number, position: number): number {
+            if (stream.flags & stream.isAppend && stream.node.size) {
```
neilccbrown: I'd maybe spell this out slightly more, something like:

    if ((stream.flags & stream.isAppend) != 0 && stream.node.size) {

I overlooked the single & for a minute and was confused because I thought you were using isAppend as a boolean, but actually it's a mask?

neilccbrown (commented):
Generally good, just a couple of small suggestions for clarification.

src/stryperuntime/pyodide-emscripten-cloud-fs.ts
PwtKCL: Yes that's why I meant by the name is confusing in some comment. It looks like it should be a boolean, but it is a mask value. OK good point for the modification, I'll change it.

PR #945, opened 2026-07-07
Various bug fixes

neilccbrown:
This fixes a few different bugs:
 - Fix an issue with media literal resizing being off by one pixel (e.g. it says 100x60, you get 99x59).
 - Fix "except Exception as e" not saving and reloading correctly.
 - Fix sounds not accepting a special library syntax, which broke the mediacomp sound examples
 - Fix an issue with newlines in project/class/function documentation fields upsetting the line number calculation, so that if you then got a runtime error it could display in the wrong place, or cause a "silent" exception in the console that made it seem like execution had hung.

### DIFF
--- a/pysrc/strype/sound.py
+++ b/pysrc/strype/sound.py
@@ -165,7 +165,8 @@ def load_sound(source):
     # If they mistakenly try to load a sound (e.g. a literal) just let it through:
     if isinstance(source, Sound):
         return source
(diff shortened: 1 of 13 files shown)


PwtKCL (commented):
It looks the tests for resizing need a tweak, but otherwise all good to merge in.

PwtKCL:
As tests are only failing because of timing issues, I merge now (we will integrate the changes in a new version very shortly).

PR #950, opened 2026-07-09
Update book projects (chapter 8)

PwtKCL:


PwtKCL:
As this is only some project changes, I'll merge them directly @mkolling

PR #961, opened 2026-07-09
Various fixes paste

PwtKCL:
This PR mainly fixes some issues around undo/redo of pasting code.
It fixes  #947, #940 and another issue when the paste over a selection was not completed (because of invalid Python): we need to restore the selection as it was.


neilccbrown:
All fine.  (Re: the test failures, I have a fix for the resizing test failures as part of a bit of work on flaky tests.  Will merge that later on after the release, though.)

PR #964, opened 2026-07-10
Handle two stores in our IndexedDB

PwtKCL:
This PR fixes #963 which was a consequence of a problem with IndexedDB.

The code was creating one DB store for a given Strype version. 
When Strype had never been been accessed on the browser (or at least not since the new model), accessing Strype from the browser was detecting that the required store was missing and created it.

That is fine, but then when accessing _the other version of Strype_, the store _of this other version of Strype_ didn't exist and couldn't be created, because the DB version was the same so we didn't have any way of creating the store (DB schema can only be changed on `onupgradeneeded`).

Therefore, if we want to have 2 stores for our DB, we need to create them both whenever we create one. The version of our IndexDB has been bumped to reflect the change and adapt to existing DB in the users' browser.


neilccbrown:
Well spotted!  Maybe eventually we should have a test for this, but anyway this makes sense to me so I think fine to deploy.

PR #965, opened 2026-07-12
Fix issues with function pasting and errors on function headers

neilccbrown:
This fixes Michael's asteroids project.  It has three changes (plus tests):
 - It fixes a bug where errors on function headers would have the wrong line number mapping, preventing the error showing.
 - It fixes a bug where pasting functions with "self" as the first param would retain it in a slot, which would then end up "doubled" with Strype's automatic self.  Instead we should just remove self when pasting, because in classes we add it already, and in top-level functions we don't want self anyway (in the case we've pasted a function copied from a class)
 - Also moved the error location tests out of console-execution (which was growing too large) into their own new file.


PwtKCL (approved):

PwtKCL:
The failing tests aren't related to the PR, so it can be merged.

PR #966, opened 2026-07-13
Fixed member method documentation not showing up for Actor

neilccbrown:
We needed to preserve the doc string when extracting the types to pass across to TigerPython (which does understand doc strings).

Fixes #952


PwtKCL (commented):
One test related to documentation parsing fails but otherwise all good.

PR #968, opened 2026-07-14
Various fixes

PwtKCL:
This PR contains simple fixes:
- allow brackets in a for frame LHS part (#935)
- do not allow the click action on "save as" when the menu entry is disabled (#967)
- keep divider persistence over some layout change - some of them were lost when switching modes, making it appear we would save them because the UI looked modified, but the state was not (#960)

The OOP project is also amended for a better layout (the old one was ignored anyway because of the bug aforementioned).


neilccbrown:
All looks fine to merge once the tests finish.

neilccbrown:
New changes look good, and fine to merge if the tests pass.

PR #969, opened 2026-07-15
Stop the tests being so flaky

neilccbrown:
This is a set of changes (authored with Claude) to try to improve our flaky test situation which had gotten very annoying: PRs would fail the tests more often than not and require multiple tries even though it was unrelated to the latest changes.  I don't think this PR makes it completely perfect, but the last two CI runs before writing this passed all tests, so I think it's definitely much better.  It's ended up a massive PR so I'll try to summarise the changes here (roughly in the order Github shows them).  My suggested review strategy is essentially: review the main code changes (if GitHub is consistent in its order, all the main code changes show before all the test code changes) and just skim the test code changes because they're quite boring and samey around changing timeouts to waiting for DOM changes.  I've written notes as I went down reading the changes, so these notes are roughly in order with the changes made in the code if you view all changes.

Playwright/CI test changes:

 - I've changed which browser+OS combinations we test with Playwright to be in line with likely usages by users.  It's now:
    - Chrome on all OSes, including Windows
    - Firefox on Ubuntu (Firefox on Windows has an odd failure, Firefox on Mac is just super slow)
    - Webkit on MacOS only (No-one uses Webkit outside Safari on MacOS, really)
 - The MacOS Playwright tests are split into two parts.  One is a single-threaded part for tests which just seemed to fail with long timeouts because they couldn't handle being run in parallel.  This includes the does-the-program-stop-after-a-tight-loop test, and the does-the-storage-model-work-with-multiple-pages-open test.  The rest are the standard parallelised tests.
 - When running Playwright, you can now use the GREP environment variable to only run tests that match that name, SPEC can have multiple comma separated items, and EXCLUDE_SPEC does the opposite of SPEC; runs everything except those specs.
 - Minor change: I've cached the Pyodide download that the Github CI performs to speed things up slightly.
 - Minor change: the Playwright test summaries on GitHub are now collapsible, and collapsed by default, making it easier to see where the failures are (a failure icon shows next to the collapsed heading if it contains a failure).

Misc changes:
 - There is a new CLAUDE.md file that Claude uses.  I've left it in the repository for future usage.  It's a bit annoying to have it but think of it like committing the VS code settings; it helps if multiple developers ever use it.
 - Microbit PYI files moved to pysrc to stop a complaint that would fill up the logs about file paths

Changes in main code (not test code) to fix actual bugs/flakiness this work uncovered:
 - Ctrl-space at frame cursor has been changed from relying on new frame being available on nextTick and instead relying on the return of an addFrameWithCommand Promise.
 - Similar change for converting frames to variable assignment; wait for the element to be in the DOM before setting cursor rather than relying on double nextTick.
 - Changed the divider setting on load from using timeouts to using a test-like mechanism of waiting until the panes stop moving.
 - Fix timing issue that was primarily seen in CI (not sure users could be this fast) where a "set focus later" could fire after a dialog was closed if the save dialog was very quickly closed.
 - Fixed an issue in the open book projects where if the user clicked on a chapter before all chapters had loaded, we would set it back to chapter 1 afterwards, which would be mildly annoying and also caused a test to fail.
  - Improved an issue with stopping the code during a tight loop especially on Safari.  It seems that Safari won't kill a process which is blocked on a communication so we need to interrupt it then kill it.
 - Not a fix but as part of the tests we expose the nextTick function in the main code so the tests can access it for some changes.
 - Improved the graphical Actor/Sprite updates to also do the catch-up mechanism we recently added for prints.  That one fixes a bug (#953).

Then the rest is test changes.  The primary idea there is to replace timeout waits (e.g. wait 500 milliseconds) with waiting for changes in the DOM (e.g. wait until this element is added).  To that end there's a couple of new helper methods (one for Cypress, one for Playwright) called waitForEditorSettled() that generically waits until the editor stops making nextTick and DOM changes.  There's a couple more methods with similar ideas: waitForProjectNameOrTimeout waits for the project name to be changed after a load, waitForGraphicsSettled waits until the graphics canvas stops redrawing, waitForConsoleSettled waits until the console stops changing.

I've also combined a lot of very similar initialisation code in each Playwright file (which all had a very similar beforeEach block) into a shared helper method (setupStrypeTest) to allow easier changes to that initialisation.

You'll see that beyond half the way in this PR it mostly devolves into just a mass replacement like this:

    -    await page.waitForTimeout(200);
    +    await waitForEditorSettled(page);

This should also be how we write it going forward; use the settled function rather than a guessed timeout.  This fits how Cypress and Playwright always tell us they should be used; waiting for DOM changes rather than waiting for timeouts.  And it does seem to help with flakiness.

I'd suggest for the last 2/3rds or 1/2 of the PR you can basically page-down, ignoring all one line diffs and only stopping for larger blocks, which are usually just changing to use the setup methods.  So most of the test code changes are uninterested, outside perhaps of the definitions of those new helper methods.

Although the tests do seem to pass I am still suspicious that two bugs in particular remain in our code:
 - I think sometimes the project name fails to set after loading.  Claude's theories on this were all incorrect and I couldn't get it to reproduce reliably.  But I just have a feeling there's still a bug there.
 - There is a fix to improve stopping the code with a tight loop in time on Safari on MacOS, which Michael saw.  But sometimes after that fix it still took a long time.  That test now runs without parallelism as parallelism seemed to make it much worse, but I have a feeling it still might not be completely resolved on Safari specifically.  We'll see.

(This summary here was totally written by me.  Claude was definitely helpful here to come up with all the settled methods and to do the mass replacement, and also to spot and fix some of the issues in the actual main code.  Even though it still took a week to do this I think manually it would have taken me several weeks to do all this, and concentrating on it more fully rather than just occasionally answering Claude.  I have reviewed all the code myself in writing this PR.  Claude counts this initiative as paused as I didn't quite replace every last timeout in the code but since the tests started passing reliably it was diminishing returns to do the rest so I thought I'd stop here; can always come back to it later.)

Fixes #953


src/helpers/editor.ts
```diff
@@ -1899,6 +1923,56 @@ export function getNumPrecedingBackslashes(content: string, cursorPos : number)
...
+        let stableSince = Date.now();
+        let lastSizes = "";
+        const check = () => {
+            if (settled) {
+                return;
+            }
+            const sizes = Array.from(document.getElementsByClassName("splitpanes__pane"))
+                .map((pane) => {
+                    const rect = pane.getBoundingClientRect();
+                    return Math.round(rect.width) + "x" + Math.round(rect.height);
```
PwtKCL, line 1957: Why do you want to round() the values here. I guess there wouldn't be much decimal variation in the size but then should we keep the values as they are ?

PwtKCL (commented):
Amazing work, well done !! (to you mainly, even if Claude was helping you've still made it happen!)
I have only a question about a code change otherwise everything seems good.
Only one test failed ("Offer to reload unsaved backups > Load several states, save some (2nd: false), then load new one") but if that's purely because of some timeout issue, it can be merged. Only one test failing is an achievement compared to what we had before.

src/helpers/editor.ts
neilccbrown: It's to prevent against very minor changes counting as not-yet-settled.  (The size here is just used for the settling check.)

Claude's longer version of an answer:   This function polls pane.getBoundingClientRect() — the raw DOM API, not Playwright's boundingBox() — inside a requestAnimationFrame loop, and declares the layout "settled" once the stringified size stays unchanged for stableMs (150ms). Raw getBoundingClientRect() returns full floating-point sub-pixel precision (e.g. 342.33333321889984), and CSS flex/percentage-based layout (which is what splitpanes uses) is prone to producing tiny sub-pixel differences between animation frames even when a pane is visually at rest — floating-point noise in the layout engine's own calculations, not a real ongoing resize. Without rounding, an exact string/float comparison across consecutive frames could very plausibly never match, so stableCount/stableSince would never accumulate and every call would fall through to the 5000ms hard timeout instead of resolving in ~150ms after a genuine settle.

PR #970, opened 2026-07-17
Add strype.graphics and strype.sound imports as default

neilccbrown:
This makes the in-theory "simple" change to add two import frames to the starting Python state.

The problem was this has a lot of knock-on effects on the tests, many of which implicitly assumed there were no imports in the starting project.  This is fixed in two different ways depending on the test:
 - Delete the imports.  There are now some helper methods to clear the imports to help with this.
 - Keep the imports.   In a few cases we leave the imports in-place but account for them being present e.g. saving the default project should have those imports.

As Claude was doing its thing, it also found an existing bug in frame deletion where deleting could sometimes cause an exception.  You can see more about that in the individual commit comment (the 3rd commit, "Fix crash when Delete has no next frame to move to").

I also tweaked a couple of flaky tests unrelated to this work: bumped one timeout and (at Claude's suggestion after an out-of-memory error) changed one long running paste test to be many short-running tests.  So I think the tests should all pass but we'll see.

Fixes #951

### DIFF
--- a/src/store/store.ts
+++ b/src/store/store.ts
@@ -2117,13 +2117,13 @@ export const useStore = defineStore("app", {
                     if(!foundDisabledJointFrameToDelete) {
                         const indexOfCurrentInAvailables = availablePositions.findIndex((e)=> e.frameId === currentFrame.id && e.caretPosition === this.currentFrame.caretPosition);
                         // the "next" position of the current
-                        frameToDelete = availablePositions[indexOfCurrentInAvailables+1]??{id:-100, isSlotNavigationPosition: false};
+                        frameToDelete = availablePositions[indexOfCurrentInAvailables+1]??{frameId:-100, isSlotNavigationPosition: false};
                     }
                     // The only times to prevent deletion with 'delete' is when we are inside a body that has no children (except in Joint frames)
                     // or when the next position is a joint root's below OR a strict* block frame below - both enabled (* = that cannot take joint frame, like a function definition, a class definition or "with")
-                    if((framesIdToDelete.length==1 && this.frameObjects[framesIdToDelete[0]]?.frameType.allowChildren && !this.frameObjects[frameToDelete.frameId]?.frameType.isJointFrame 
+                    if((framesIdToDelete.length==1 && this.frameObjects[framesIdToDelete[0]]?.frameType.allowChildren && !this.frameObjects[frameToDelete.frameId]?.frameType.isJointFrame
                             && this.currentFrame.caretPosition == CaretPosition.body && this.frameObjects[framesIdToDelete[0]]?.childrenIds.length == 0)
-                        || (!this.frameObjects[frameToDelete.frameId].isDisabled && (!!this.frameObjects[frameToDelete.frameId]?.frameType.allowJointChildren || (this.frameObjects[frameToDelete.frameId]?.frameType.allowChildren && !this.frameObjects[frameToDelete.frameId]?.frameType.allowJointChildren))
+                        || (!this.frameObjects[frameToDelete.frameId]?.isDisabled && (!!this.frameObjects[frameToDelete.frameId]?.frameType.allowJointChildren || (this.frameObjects[frameToDelete.frameId]?.frameType.allowChildren && !this.frameObjects[frameToDelete.frameId]?.frameType.allowJointChildren))
                             && (frameToDelete.caretPosition??"") === CaretPosition.below)){
                         frameToDelete.frameId = -100;
                     }
--- a/tests/cypress/support/test-support.ts
+++ b/tests/cypress/support/test-support.ts
@@ -37,14 +37,17 @@ export function navigateToTopOfCode(): void {
 // for the test's own import to be typed/pasted.
 export function clearDefaultImports(): void {
     navigateToTopOfCode();
-    cy.then(() => deleteFramesUpTo("#frameContainer_-1", Cypress.$("#frameContainer_-1 .frame-div").length));
+    cy.then(() => deleteFramesUpTo("#frameContainer_-1"));
 }
 
-// Presses Delete (forward) against `containerSelector`'s frame count, stopping as soon as the
-// container is actually empty rather than always pressing `maxPresses` times. `maxPresses` is
-// expected to be a safe upper bound (a count that includes nested descendants, e.g. the statements
-// inside a function/class definition's body), not an exact number of presses needed -- deleting a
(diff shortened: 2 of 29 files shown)


tests/playwright/e2e/fileIO.spec.ts
```diff
@@ -27,7 +23,13 @@ with open("/tmp/${copyFileName}","wb")  as f2  :
 `;
     return async (page) => {
         await deleteDefaultProject()(page);
-        await doPagePaste(page, createAssetCopyCode);        
+        // deleteDefaultProject leaves the caret at the top of Imports; move to Main so this paste
+        // (and the test's own paste that immediately follows, with no navigation in between) both
+        // land in Main in the right order -- the test's code reads the file this writes, so it
+        // must run second:
+        await page.keyboard.press("ArrowDown");
+        await page.keyboard.press("ArrowDown");
```
PwtKCL: Not a big deal, but we can use the method you wrote `pressN` to make it simpler to read. Up to you.

tests/playwright/e2e/load-save-demos-books.spec.ts
```diff
@@ -45,8 +45,13 @@ test.describe("Load/save book projects", () => {
...
             const original = readFileSync("public/book_projects/chapter02/fireworks.spy", "utf8").replace(/\r\n/g, "\n");
-            await page.keyboard.press("Delete");
-            await page.keyboard.press("Delete");
+            await clearDefaultProject(page);
+            // clearDefaultProject leaves the caret at the top of (now empty) Imports; return to
+            // Main so the "move up i times" logic below still exercises pasting with the caret in
+            // Main/Defs/Imports respectively, now that Imports itself has also been cleared of its
+            // default content:
+            await page.keyboard.press("ArrowDown");
+            await page.keyboard.press("ArrowDown");
```
PwtKCL: Same about `pressN`

tests/playwright/e2e/scroll-into-view.spec.ts
```diff
@@ -136,8 +136,11 @@ test.describe("Undo scrolls location into view", () => {
     ];
     for (let testIndex = 0; testIndex < undoTests.length; testIndex++) {
         test(`Undo test #${testIndex}`, async ({page}) => {
-            await page.keyboard.press("Delete");
-            await page.keyboard.press("Delete");
+            await clearDefaultProject(page);
+            // clearDefaultProject leaves the caret at the top of Imports; return to Main, which
+            // is where this test's content belongs:
+            await page.keyboard.press("ArrowDown");
+            await page.keyboard.press("ArrowDown");
```
PwtKCL: Same about using `pressN`

tests/playwright/e2e/scroll-into-view.spec.ts
```diff
@@ -201,8 +204,11 @@ test.describe("Undo scrolls location into view", () => {
 
 test.describe("Printing scrolls into view", () => {
     test("Pressing keys repeatedly scrolls into view", async ({page}) => {
-        await page.keyboard.press("Delete");
-        await page.keyboard.press("Delete");
+        await clearDefaultProject(page);
+        // clearDefaultProject leaves the caret at the top of Imports; return to Main, which is
+        // where this test's content belongs:
+        await page.keyboard.press("ArrowDown");
+        await page.keyboard.press("ArrowDown");
```
PwtKCL: Same about using `pressN`

tests/cypress/support/test-support.ts
```diff
@@ -24,9 +24,60 @@ export function focusEditorAndClear(): void {
...
+        for (let i = 0; i < mainCount; i++) {
+            cy.get("body").type("{del}");
+            waitForEditorSettled();
+        }
+    });
+    cy.get("body").type("{uparrow}");
+    waitForEditorSettled();
+    cy.then(() => {
+        const defsCount = Cypress.$("#frameContainer_-2 .frame-div").length;
+        for (let i = 0; i < defsCount; i++) {
```
PwtKCL: Here and everywhere we have a loop doing "press" then "wait", we can use the method `pressN` you had written, and drop the for loop.

src/store/store.ts
```diff
@@ -2117,13 +2117,13 @@ export const useStore = defineStore("app", {
                     if(!foundDisabledJointFrameToDelete) {
                         const indexOfCurrentInAvailables = availablePositions.findIndex((e)=> e.frameId === currentFrame.id && e.caretPosition === this.currentFrame.caretPosition);
                         // the "next" position of the current
-                        frameToDelete = availablePositions[indexOfCurrentInAvailables+1]??{id:-100, isSlotNavigationPosition: false};
+                        frameToDelete = availablePositions[indexOfCurrentInAvailables+1]??{frameId:-100, isSlotNavigationPosition: false};
```
PwtKCL, line 2120: Good catch! I'm a bit surprised that the TS language server doesn't pick it up actually.

tests/playwright/support/editor.ts
```diff
@@ -296,34 +315,34 @@ export function getDefaultStrypeProjectDocumentationFullLine(): string {
...
+    // pressing it more times than could possibly be needed is harmless -- but a block frame's body
+    // is itself an extra caret stop beyond the one .frame-div element it counts as, so totalCount
+    // alone can undershoot for nested content; multiplying it gives enough headroom for that
+    // without needing to know the exact nesting depth:
+    for (let i = 0; i < totalCount * 3 + 10; i++) {
+        await page.keyboard.press("ArrowUp");
     }
-    // Navigate up one level at a time, deleting backward to clear Definitions then Imports:
-    await page.keyboard.press("ArrowUp");
     await waitForEditorSettled(page);
```
PwtKCL, line 343: Either this call should be inside the for loop, or we can use `pressN` as suggested before.

tests/playwright/e2e/rename-identifiers.spec.ts
PwtKCL, line 26: I suppose we can add imports as we did for the `initialProjectCodeLiteral` code constant above, and simplify this test file.

PwtKCL (commented):
I have some suggestion for discarding the for loops if we wanted to and one wait helper function that seemed misplaced.
Otherwise, I think at least one Playwright test (rename-identifers) can be simplfied by only updated the constants - if I'm not mistaken.
Overall though it seems fine. Let's see the tests completion.

neilccbrown:
I've used pressN as you suggested and added a comment about the other line, so then I will merge if the tests pass.

PR #971, opened 2026-07-18
Fix sound saving and a flaky test

neilccbrown:
Two small fixes from yesterday which I'm filing while I remember what was done, but not urgent, just look at it next week.  Fixes:
 - Fix the bug Michael found with saving [soundliteral].play() sometimes having ___strype_blank in it.  Turned out to just be an indexing issue in one of the slot handlers.  Added a test for this (although it was an intermittent issue but we can watch in case this test ever fails again.)
 - The first test run failed with something I'd seen before where testing opening the fireworks project failed.  That turned out to be a race hazard between loading the dialog and the user clicking; if they clicked quick enough the dialog would later "initialise" to the first chapter and set it back.  Maybe the user would have just been mildly annoyed but it broke the test which was "click on second chapter, click on project".  So I've fixed to do the initialisation earlier.  See what you think about adding the show method; it seemed a sensible fix but if you're not keen we can find a different way.

### DIFF
--- a/src/components/Menu.vue
+++ b/src/components/Menu.vue
@@ -422,8 +422,9 @@ export default defineComponent({
         );
 
         // The events from Bootstrap modal are registered on eventBus.
+        eventBus.on(CustomEventTypes.strypeModalShow, this.onStrypeMenuShowModalDlg);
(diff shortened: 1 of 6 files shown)


PwtKCL (approved):
All good. I think using the show event is the right solution.

PR #973, opened 2026-07-19
Add dynamic operator spacing

neilccbrown:
I was looking to fix the bug we discussed last meeting where spacing around the keywords operators was too tight.  But then I thought while I was fixing that I'd get the AI to implement Stride's dynamic operator spacing, where it picks the spacing based on the expression dynamically.  So for example 1+2 gets default spacing, then 1+2*3 has tighter spacing around the multiply than the addition, and if you have a third tier (** does it I think) then it changes again.  With the largest range of operators the symbol-operator spacing comes down to zero, but that doesn't suit keyword operators (and in general, keyword operators want a bit more spacing around them) so keyword operators have their own tier.  There's also a separate part of the system for unary operators (not, ~, +, -) which should not have left-hand spacing.

As part of this, also discovered that unary ~ and "lambda" keyword could not be pasted into Strype, so fixed that.  If you want to see what it looks like, spin up a dev server from this branch and try loading the file tests/manual/operator-spacing-demo.py which has a variety of operator expressions for you to be able to see what it's like.  The spacing should change as you add/delete operators if they are different precedences.  I wonder if the keywords have slightly too much space (especially when next to a blank slot which matters for us internally but isn't obvious to the user (e.g. in "(a) and (b)" there's a blank slot either side of the "and")  but if we're happy with the principle of the system we can tweak exact spacing later on.

Sorry for another PR; I think this is the downside of AI, more PRs to review...  But it's not really as big as it looks on the main code side, because a lot of the change is a big new test file.

### DIFF
--- a/src/components/LabelSlot.vue
+++ b/src/components/LabelSlot.vue
@@ -92,6 +92,7 @@ import { isMacOSPlatform } from "@/helpers/common";
 import { vueComponentsAPIHandler } from "@/helpers/vueComponentAPI";
 import { eventBus, projectDocumentationFrameId } from "@/helpers/appContext";
 import { detectBrowser } from "@/helpers/browser";
+import { PrecedenceTier, UNARY_PREFIX_OPERATORS } from "@/helpers/operatorPrecedence";
 
 // Default time to keep in cache: 5 minutes.
 const soundPreviewImages = new Cache<LoadedMedia>({ defaultTtl: 5 * 60 * 1000 });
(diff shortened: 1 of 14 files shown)


src/components/LabelSlot.vue
```diff
@@ -147,6 +148,10 @@ export default defineComponent({
         isEditableSlot: Boolean,
         isEmphasised: Boolean,
         isFrozen: Boolean,
+        precedenceTier: {
```
PwtKCL: I'd just rename this property "opPrecedenceTier" for clarity.

PwtKCL (approved):

src/components/LabelSlot.vue
neilccbrown: Done.  I will merge if/when the tests pass.

Issue #757, opened 2026-03-04
Python code with blank in joint frame fails to paste


neilccbrown:
I was trying to paste code like this:

        if os.path.isdir(full_path):
            print(" " * indent + f"[{entry}]")
            pretty_print_directory_tree(full_path, indent + 4)

        elif os.path.isfile(full_path):
            print(" " * indent + entry)

And it failed until I took the blank line out before the elif.  We should allow this to parse and paste.

Issue #543, opened 2025-07-04
Pasting raw Python in an if/elif structure can be faulty


PwtKCL:
(Note: this needs to be checked when the fix to #542 is available as we need to be able to paste inside a frame)
Consider the following Strype code:
![Image](https://github.com/user-attachments/assets/40b947f2-f85f-48b3-88b6-9a4285c5e40f)

And the following Python code to copy:
```
else :
   print("a")
```

The following will happen:
- frame cursos is above `a()`: wrong paste
![Image](https://github.com/user-attachments/assets/30be4eae-a8d9-469f-95ee-84702e233c3a)

- frame cursor is below `a()`: right paste
- frame cursor is above `b()`: wrong paste
![Image](https://github.com/user-attachments/assets/d1d80cfd-f2c6-45c7-8945-0f41a405eb14)

- frame cursor is below `b()`: right paste

Regardless the output is correct or not, however, the frame cursor gets lost.

PR #974, opened 2026-07-20
Fix a few paste bugs

neilccbrown:
This fixes a few paste issues, split by commit so might be easiest to review individual commits:
 - Support pasting/loading "a += b" by turning it into "a = a + b".  Brackets are added if b is not a single-item expression (so if you paste "x += 3 * 5" it comes out to "x = x + (3 * 5)".
 - Fix an issue where pasting a blank line in the middle of an if/else (before the else) was causing an exception, when really we should accept it (see https://github.com/k-pet-group/Strype/issues/757)
 - Fixed an issue where pasting else above elif could cause an invalid state (see https://github.com/k-pet-group/Strype/issues/543)
 - In the course of doing this, the AI spotted an invalid state we could get into where the cursor position was below an if despite having joint frames following (this cursor position sort of exists in our minds, but it shouldn't be reachable by the user) and pastes there would fail.  So it fixed that in the last commit.

### DIFF
--- a/src/store/store.ts
+++ b/src/store/store.ts
@@ -2656,19 +2656,33 @@ export const useStore = defineStore("app", {
             const areFramesJoint = payload.sourceFrames.frames[payload.sourceFrames.frameIds[0]].frameType.isJointFrame;
             if (areFramesJoint) {
                 // Joint frames.  We can only paste if we're inside the body of the joint parent
-                // or one of its joint frames, at the last position of the body:
-                let parentId: number;
-                if (payload.target.caretPosition == CaretPosition.body) {
-                    parentId = payload.target.id;
+                // or one of its joint frames, at the last position of the body -- or immediately
+                // below the joint parent's own frame, which is where the caret naturally ends up
+                // once nothing follows its whole chain (e.g. pasting a trailing "else" right after
+                // finishing an if/elif with nothing else in the body).  That's the joint parent's
+                // own "below" position, not one of its children's, so it needs to append after any
+                // existing joint children rather than being treated as "top of its body":
+                let jointParentId: number;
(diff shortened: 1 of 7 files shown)


src/store/store.ts
```diff
@@ -2656,19 +2656,33 @@ export const useStore = defineStore("app", {
...
-                // or one of its joint frames, at the last position of the body:
-                let parentId: number;
-                if (payload.target.caretPosition == CaretPosition.body) {
-                    parentId = payload.target.id;
+                // or one of its joint frames, at the last position of the body -- or immediately
+                // below the joint parent's own frame, which is where the caret naturally ends up
+                // once nothing follows its whole chain (e.g. pasting a trailing "else" right after
+                // finishing an if/elif with nothing else in the body).  That's the joint parent's
+                // own "below" position, not one of its children's, so it needs to append after any
+                // existing joint children rather than being treated as "top of its body":
```
PwtKCL, line 2664: I understand why we do it, but I don't know if we should allow it nonetheless. It means we have a dual behaviour in Strype regarding adding something to a joint frame: either pasting at its below position, or adding via the usual frame UI inside its body.

PwtKCL (commented):
Just one command about pasting below the joint structures, otherwise it seems all good. Let's see how tests pass.

src/store/store.ts
neilccbrown: I think I will leave it in place for now as it's working but something to consider for future rearrangement.  I also just made one further small modification as we had a test which expected "-=" to fail parsing but now it doesn't, so adjusted it (add a different invalid operator, "@") so it does fail parsing again as expected.

PR #976, opened 2026-07-20
Show shortcuts while text editing

neilccbrown:
This is a small PR which adds a feature I'd wanted for a while: showing the Ctrl-Space shortcut while editing a slot in the commands pane so that it's discoverable.  It's not actually clickable (it would remove focus from the text slot so it would be fiddly to restore focus then show the dialog).  Also when you have a selection in slots, adds the Wrap-in-brackets items as shortcuts, as they are also not really discoverable at the moment.

This is a precursor to a future planned change; I want to add the ability to insert microphone recordings as sound literals and webcam images as media literals.  Having a menu item for them is awkward so I think a keyboard shortcut is suitable (Ctrl-Shift-M or whatever we pick) and this way it could actually be discoverable.

### DIFF
--- a/src/localisation/en/en_main.json
+++ b/src/localisation/en/en_main.json
@@ -249,11 +249,11 @@
     "codeCompletion": "Code completion",
     "codeCompletionFilePaths": "Code completion (file paths)",
     "spaceKey": "space",
-    "wrapRoundBrackets": "Wrap in round brackets",
-    "wrapSquareBrackets": "Wrap in square brackets",
-    "wrapCurlyBrackets": "Wrap in curly brackets",
-    "wrapDoubleQuotes": "Wrap in double quotes",
-    "wrapSingleQuotes": "Wrap in single quotes"
+    "wrapRoundBrackets": "Wrap in ()",
+    "wrapSquareBrackets": "Wrap in []",
+    "wrapCurlyBrackets": "Wrap in {'{}'}",
+    "wrapDoubleQuotes": "Wrap in \"",
+    "wrapSingleQuotes": "Wrap in '"
   },
   "commandTabs": {
     "0": "Add Frame"
--- a/src/localisation/en/en_main.json
+++ b/src/localisation/en/en_main.json
@@ -245,7 +245,10 @@
     "importedModules": "Imported modules",
     "builtinFiles": "Built-in files",
     "invalidState": "No completion available",
-    "noDocumentation": "No documentation available"
+    "noDocumentation": "No documentation available",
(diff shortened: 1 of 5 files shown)


src/localisation/en/en_main.json
```diff
@@ -245,7 +245,15 @@
     "importedModules": "Imported modules",
     "builtinFiles": "Built-in files",
     "invalidState": "No completion available",
-    "noDocumentation": "No documentation available"
+    "noDocumentation": "No documentation available",
+    "codeCompletion": "Code completion",
+    "codeCompletionFilePaths": "Code completion (file paths)",
```
PwtKCL, line 250: I don't know if there is a clearer explanation possible? "file paths" may be unclear (although it's getting obvious once the list shows..)

src/localisation/en/en_main.json
```diff
@@ -245,7 +245,15 @@
...
+    "spaceKey": "space",
+    "wrapRoundBrackets": "Wrap in round brackets",
+    "wrapSquareBrackets": "Wrap in square brackets",
+    "wrapCurlyBrackets": "Wrap in curly brackets",
+    "wrapDoubleQuotes": "Wrap in double quotes",
+    "wrapSingleQuotes": "Wrap in single quotes"
```
PwtKCL: I'd suggest a straight forward symbolic explanation for these, like "Wrap with **(** **)**" (the bold here would be the same in the UI) - in English the literal names are straight forward definitions, but in French and maybe other languages they're not (and I'm not sure they will be obvious for younger users)

src/localisation/fr/fr_main.json
```diff
@@ -245,7 +245,15 @@
     "importedModules": "Modules importés",
     "builtinFiles": "Fichiers intégrés",
     "invalidState": "Aucune complétion disponible",
-    "noDocumentation": "Aucune documentation disponible"
+    "noDocumentation": "Aucune documentation disponible",
+    "codeCompletion": "Complétion de code",
+    "codeCompletionFilePaths": "Complétion de code (chemins de fichiers)",
```
PwtKCL: "chemins d'accès de fichier" is more accurate here :) but again it may be quite obscure when users didn't see the list content. I'm not sure if "fichier" should be singular or plural actually, but I'd say singular it's several paths for a file, not several paths for several files.

PwtKCL (commented):
Just a few comments about the wordings, but otherwise all fine. I

neilccbrown:
I've changed the text in the labels as you suggested, although couldn't think of a better name for the file paths.

PR #980, opened 2026-07-23
Fix bugs with match and console input

neilccbrown:
This fixes the two bugs @mkolling reported today:
 - The match statement's default case had an error.  The AI pinpointed this as being an issue with having a value for slot index 1 for the default case frame even though there's only one case slot (index 0).  Added a test and a fix for this.
 - The console input bug was already filed as #927 and there was a suggestion in there on how to fix, so that's not been done.  If you stop the program during console input, it's now cleared up properly so that if you re-run you can enter input on the second run.

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -797,7 +797,7 @@ export function generateAllFrameDefinitionTypes(regenerateExistingFrames?: boole
         colour: "#E0DFE4",
         forbiddenChildrenTypes: Object.values(AllFrameTypesIdentifier).filter((type) => type != StandardFrameTypesIdentifiers.case && type != StandardFrameTypesIdentifiers.blank && type != StandardFrameTypesIdentifiers.comment),
         // A match statement must always have one case at least, so we enforce it upon frame creation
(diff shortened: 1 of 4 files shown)


neilccbrown:
I briefly added an extra commit about mouse positions but I've realised that needs more work so I've put it back to just the commits mentioned in the PR description.  (Hence the force push being mentioned in the timeline above.)

src/types/types.ts
```diff
@@ -797,7 +797,7 @@ export function generateAllFrameDefinitionTypes(regenerateExistingFrames?: boole
         colour: "#E0DFE4",
         forbiddenChildrenTypes: Object.values(AllFrameTypesIdentifier).filter((type) => type != StandardFrameTypesIdentifiers.case && type != StandardFrameTypesIdentifiers.blank && type != StandardFrameTypesIdentifiers.comment),
         // A match statement must always have one case at least, so we enforce it upon frame creation
-        defaultChildrenTypes: [{...EmptyFrameObject, frameType: CaseDefinition, labelSlotsDict: {0: {slotStructures:{fields:[{code:"_"}], operators: []}}, 1: {slotStructures:{fields:[{code:""}], operators: []}}}}],
+        defaultChildrenTypes: [{...EmptyFrameObject, frameType: CaseDefinition, labelSlotsDict: {0: {slotStructures:{fields:[{code:"_"}], operators: []}}}}],
```
PwtKCL, line 800: oops sorry. Most likely a rogue copy/paste...

PwtKCL (approved):
Since the previous tests iteration succeeded, I merge it now.

PR #982, opened 2026-07-24
Copy and download console 

PwtKCL:
This PR deals with the PEA console, fixing a few issues we had with it's initial state and restricting the drag and drop, and mainly adding some features for copying and downloading the PEA console.

- Copying the console fixes #930: via the context menu on the console or via the small icon-button located at the top-right corner of the console as per the image below. If the a text selection in the console extists, that selection is copied - otherwise the console's full content is copied.
imageimage

- Downloading the console screenshot: via a context menu (see image above). Note that the download action triggers a progress bar. In most cases, the textarea contains text of a decent length, but when it doesn't, the mechanism to generate the image for the screenshot can be slow (because we need to make a mirror div then a canvas element).

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -449,6 +449,17 @@ export default defineComponent({
                 return;
             }
 
+            // #v-ifdef STRYPE_PLATFORM == VITE_STANDARD_PYTHON_MODE
+            // Handle the keyboard shortcuts for the PEA console copy (when the PEA console is visible).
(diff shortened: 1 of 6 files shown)


neilccbrown:
Screenshots in HTML are such a pain.  They look doable and then it always seems to be a nightmare, combined with clipboard permissions...  All looks fine.

PR #984, opened 2026-07-24
Fix mouse coordinates

neilccbrown:
As mentioned in our last meeting, Strype was allowing invalid mouse positions (i.e. outside the world) to be reported, specifically the position (-400,-300) could be reported even though (-399, -299) is the bottom left corner.  Happy to let the AI do this one as it's so fiddly...  but I have tested it and it all seems to be correct.


PwtKCL (approved):
Seems all fine. Let's wait for the tests to finish, and merge when they've passed.

PR #989, opened 2026-07-27
Fix some load recent issues

neilccbrown:
I had a poke around and found a few cases where load-recent wasn't working as intended.  The design (as confirmed by some of the comments) was meant to be that even if the user clicked "Discard changes" on a dialog we showed, we intended to still keep the project in case they mis-clicked, ready to open from the "Load recent" menu.  However, this didn't work correctly because we set the flags such that the project state was not seen as modified and the check for "save unmodified project" didn't fire because it looked unmodified.

So we have to do things slightly different.  This includes the cloud conflict case where they can say "Discard my local edits, load remote version"; these edits are also now retained in the load recent menu, named ".... (local edits)".

Tests have also been added for this, and the existing tests have been slightly improved to try to make them less flaky.


PwtKCL (approved):
Again, let's wait for the tests to complete. Fine to merge if they do.

PR #991, opened 2026-07-28
Fix sporadic paste-python.cy.ts message-banner failure (clear IndexedDB between Cypress tests)

neilccbrown:
## Summary

`tests/cypress/e2e/paste-python.cy.ts` occasionally failed with:

```
AssertionError: Timed out retrying after 4000ms: Expected <div.message-banner-container>
not to exist in the DOM, but it was continuously found.
```

## Root cause

Cypress reuses one long-lived browser tab across all tests in a spec file (unlike Playwright, which gets a fresh browser context per test). `standardBeforeEach` clears `localStorage`/`sessionStorage` between tests, but not IndexedDB.

Strype persists session-recovery state to IndexedDB. Any test that goes through the "discard unsaved changes, then load a new project" flow — e.g. `testRoundTripImportAndDownload`/`loadFile`, both used earlier in this same spec file — writes a `stillAlive: false` record there (`App.vue` / `store-db-storage.ts`). `App.vue`'s startup check for "is there a recent state to restore?" only requires that record to be under 2 minutes old (`store-db-storage.ts`'s `checkForRecentSaveStates`), and isn't scoped to the tab that wrote it. Since Cypress tests run in rapid succession and IndexedDB isn't cleared between them, a completely unrelated *later* test's fresh page load can find that leftover record and pop an undismissed "restore your session?" banner mid-test — exactly the failure seen in the unrelated "Python paste errors" tests.

## Fix

Extend `standardBeforeEach` (`tests/cypress/support/standard-setup.ts`) to also clear the session-state IndexedDB store, mirroring the existing localStorage/sessionStorage clearing. Opens the DB without a version and clears whatever object stores already exist, so it doesn't need updating if the app's own DB schema/version changes.

(The DB name is a hardcoded constant rather than imported from `src/helpers/editor.ts`, since that file pulls in Vite-only asset imports — Python stubs, SCSS modules — that Cypress's own webpack preprocessor can't parse; a comment cross-references the source of truth.)

## Test plan

- [x] `npm run type:check` passes
- [x] Ran the full `paste-python.cy.ts` spec locally (85 tests): all "Python paste errors" tests passed, no message-banner recurrence. One unrelated, isolated `cy.readFile` download-timing flake occurred on a single round-trip test (file arrived slightly after the 4s read timeout) — a different, pre-existing category of flake, not this bug.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


PwtKCL (approved):
Let's see how tests go, but good to merge if they pass.

PR #1005, opened 2026-07-31
Fix two small bugs

neilccbrown:
Fixes two small bugs recently found:
 - Actors should be created directly at their initial position, not set to 0,0 then later moved.
 - Type inference should be support for list.pop().  This was fixed upstream in TigerPython, for now we point to the updated branch which includes this fix.  Thus the code change for that is just to add a test for it.

Fixes #1003 
Fixes #1004


PwtKCL (approved):
Good to merge. I'll refresh the testing versions accordingly.

PR #1007, opened 2026-08-03
Change frame insertion keys to need a prefix key, and allow typing and conversion at the frame cursor

### LINKED ISSUE #297: Support "F" shortcut for function definition frame addition
In GitLab by @PwtKCL on Sep 24, 2024, 11:08

Check if that is not creating conflicts (should be as "for" can never be shown simultaneously with function definition frames).

To confirm: "f" can be added as a hidden shortcut?

(Neil edited: f is the shortcut at the moment, the question is about adding "d", I think?)

neilccbrown:
This is the big change recently discussed to add a prefix key before frame insertion, and make typing at the frame cursor starting typing in a "generic" (secretly: funccall without trailing brackets) frame.  If the user types a keyword it is converted to that frame type.  Here's the summary of changes:

  - Replaced single-keystroke frame insertion (typing a bare letter like i, f, w to create if/for/while frames) with a new mechanism: pressing Tab or Space at the frame cursor shows a blue rectangle around the command pane.
  - `#` and Enter still insert comments and blanks directly, without needing a prefix (although you can use the prefix if you want)
  - Typing any other character now starts a func-call frame without the brackets instead.  Then:
      - Typing a recognized keyword there (e.g. if , while , return, elif/else/except/finally) converts that in-progress func-call frame into the proper keyword frame instead.
        - There are various triggers allowed after the keyword.  Space is the obvious one for if and while, but it doesn't make sense on some frames:
           - For try and else colon also works (because it's more natural to type "else:" than "else :" or "else ").
           - For things like continue, break, etc, it also doesn't make sense to press space after, so return works if there is no trailing slot.
           - For the "return" keyword there is a difference in behaviour: "return<space>" moves to the optional slot after it, but "return<enter>" moves to the frame cursor below.  Similarly for "except", "except<space>" moves to the optional slot, whereas "except:" and "except<enter>" move to the frame cursor inside the body.
        - elif/else/except/finally convert into joint frames.  Additionally, typing them just after their joint parent also converts them (without this, it's very natural to add "else:" after an "if:" frame's end, and it does nothing but give an error).  If there's ambiguity (e.g. if inside if) it associates like the joint frames, to the one we are inside.
  - "d" was made the official function-definition shortcut, with "f" kept working as a secret/legacy alias (Michael's request, and was on our list for ages as #297).

My suggestion is to almost completely ignore the test code, as that is some new tests, but has huge amounts of changes throughout to press more keys in order to get the frame insertions to work.  I changed some tests that were inserting function calls to just type without a prefix key, so overall we should have a bit of a mix in the tests, which I think is a good thing long term to check for issues.  I did repeatedly ask the AI to simplify where possible but I think ultimately: it's a huge change, so it's a large PR...

I think the tests should be passing but even if there's one or two failures to fix, the vast bulk of the code will no longer change and can be reviewed.
     
Fixes #297


PwtKCL (approved):
We will discuss further about using "tab" in the new mechanism, but otherwise the new functionality is great and the code seems fine.
One test failed, but I think we can investigate later, let's see anyway what the tests started by the merge will do.
Well done :)

PR #1008, opened 2026-08-04
A couple of small fixes to the open dialog

neilccbrown:
Two small fixes to the open dialog, requested by Michael:
 - Restyle the examples/book items
 - Remove the open recent projects menu as it was cluttering up the dialog.  I decided to actually put it on a hidden shortcut (Ctrl-U).  We get emails to BlueJ support saying "my project is lost, help" so it seems like maybe it might be useful to have this menu available if it's needed by someone who gets in touch.

### DIFF
--- a/tests/playwright/e2e/storage-model.spec.ts
+++ b/tests/playwright/e2e/storage-model.spec.ts
@@ -377,9 +377,25 @@ async function assertRecentStatesShowing(page: Page, expectedProjectNames: RegEx
     await expect(page.locator("." + scssVars.projectRecentStateLabel)).toHaveText(expectedProjectNames);
 }
 
+// The "recent unsaved projects" pane is hidden by default in the load dialog; Ctrl+U reveals it.
+// The shortcut is only armed once the dialog has fully finished its "shown" transition (a Bootstrap
+// event that fires a little after the dialog becomes visible to Playwright), so a single press right
+// after opening the dialog can race that transition. Presses before the dialog is armed are harmless
+// no-ops, so we just retry the press until the pane shows up:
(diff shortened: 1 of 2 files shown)


tests/playwright/e2e/storage-model.spec.ts
```diff
@@ -377,9 +377,25 @@ async function assertRecentStatesShowing(page: Page, expectedProjectNames: RegEx
...
 
+// The "recent unsaved projects" pane is hidden by default in the load dialog; Ctrl+U reveals it.
+// The shortcut is only armed once the dialog has fully finished its "shown" transition (a Bootstrap
+// event that fires a little after the dialog becomes visible to Playwright), so a single press right
+// after opening the dialog can race that transition. Presses before the dialog is armed are harmless
+// no-ops, so we just retry the press until the pane shows up:
+async function revealRecentUnsavedPane(page: Page) : Promise<void> {
+    const scssVars = await page.evaluate(() => (window as any)["StrypeSCSSVarsGlobals"]);
+    await expect(async () => {
+        await page.keyboard.press("Control+u");
```
PwtKCL: Should the test here check Meta for macOS?

PwtKCL (commented):
I have one small comment in the test, even if the test passes as such (I think because we are most of the times permissive enough to allow the control key for macOS as well).

tests/playwright/e2e/storage-model.spec.ts
neilccbrown: Yes good point.  I've adjusted it, will merge if the relevant tests pass.

PR #1009, opened 2026-08-04
Improve overtyping

neilccbrown:
Now that we're encouraging more keyboard typing of constructs, we want to make sure that we can overtype elements to continue with the keyboard typing approach, so this small PR:
 - Makes sure that if you type a colon at the end of the top level of a slot before a ":" label, focus moves past it (as if right arrow was pressed)
 - Similar for close-bracket in funcdef parameters
 - If you then try to type a colon at the start of the funcdef docs (e.g. you typed "foo():", the close bracket takes you into the docs, then you'd type a colon in the docs) it's discarded.  Can't think you'd want to start the docs with a colon...
 - Pressing space in the first slot of "for" or "with" moves to the second slot, like it already did for assignment frames.
 - Add tests for some of this

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -16,6 +16,19 @@ London. See [README.md](README.md) for build/run basics.
 - Single Cypress file: open the Cypress runner (`npx cypress open`) against
   the already-running dev server, or filter via `cypress-split` env vars.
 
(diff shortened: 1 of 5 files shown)


PwtKCL (commented):
Seems fine. I've looked the details of the tests, but the failed tests might be some few adjustments needed from the PR's change?

PR #1015, opened 2026-08-04
Fix loading-overlay hang and lambda blank-insertion bugs

neilccbrown:
## Summary
- Fix the "Loading file" overlay staying visible forever when a Google-Drive-shared project fails with a Python parse error. `setStateFromPythonFile()`'s promise only ever resolved (and only on success); on a parse failure it left the promise permanently pending, so callers chaining `.finally()` off it (e.g. the share-link loader) never dismissed the overlay.
- Fix a save-time bug where a call like `foo(lambda x: x)` was saved as `foo(___strype_blank lambda x: x)`. `generateFlatSlotBases()` excluded `"not"`/`"~"` from needing a left operand (both always-unary), but not `"lambda"` (also always-unary in Strype's grammar), so the blank field before `lambda` was wrongly treated as a missing binary-operator operand.

## Test plan
- [x] `npm run type:check` passes
- [x] Added a Playwright save/reload round-trip test (`"Valid lambda"` in `load-save-frame-content.spec.ts`); verified it fails against the pre-fix code (saved content: `foo(___strype_blank lambda x:x)`, reload fails) and passes with the fix.


PwtKCL (approved):
All seems good. We can merge, let's just wait for the tests to finish to make sure.

PR #1050, opened 2026-08-10
Colour picker: hex field in tile view, fix shade-tile selection

neilccbrown:
## Summary
- The colour picker's hex text field is now always visible (previously only shown in the fine-grained selector), and editing it re-runs the same "just opened" logic: an exact tile match jumps to that tile in the tinker view, otherwise it falls through to the fine-grained selector.
- Fixed a bug where picking a colour family never highlighted a shade tile: the default sat/light was the family range's raw midpoint, which doesn't land on any actual grid cell (rows/columns sit at fixed fractions of the range). Now it snaps to the real cell closest in RGB terms to the colour shown at the top of the tinker view (the selected hue chip, or the family swatch for hue-less families like grey).

## Test plan
- [x] `vue-tsc --noEmit` passes
- [ ] Manually open the colour picker (Ctrl+Shift+Y), click through several families and confirm a shade tile is highlighted each time
- [ ] Type a hex value while in tile view and confirm it switches views/selects correctly
- [ ] Run `npm run test:playwright` for `colour-picker.spec.ts`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/components/ColourPickerDlg.vue
+++ b/src/components/ColourPickerDlg.vue
@@ -102,7 +102,13 @@ import { defineComponent, nextTick, PropType } from "vue";
 import ModalDlg from "@/components/ModalDlg.vue";
 import { vueComponentsAPIHandler } from "@/helpers/vueComponentAPI";
 import { eventBus } from "@/helpers/appContext";
-import { CustomEventTypes, getPEAGraphicsContainerDivId } from "@/helpers/editor";
+import { CustomEventTypes } from "@/helpers/editor";
+// #v-ifdef STRYPE_PLATFORM == VITE_STANDARD_PYTHON_MODE
+// getPEAGraphicsContainerDivId() only exists in standard mode -- micro:bit mode has no "Graphics"
+// PEA tab/canvas for the live colour preview to dim a hole over (see punchBackdropHoleOverGraphics
+// below), so this whole live-preview mechanism is standard-mode-only.
+import { getPEAGraphicsContainerDivId } from "@/helpers/editor";
+// #v-endif
 import { BvTriggerableEvent } from "bootstrap-vue-next";
 import {
     parseColourToHex, hexToHsv, hsvToHex, hexToHsl, hslToHex,
@@ -520,6 +526,7 @@ export default defineComponent({
         // the net result is a plain rectangle with a rectangular hole, without needing an evenodd
         // fill rule (which the plain CSS polygon() function doesn't support).
         punchBackdropHoleOverGraphics() {
+            // #v-ifdef STRYPE_PLATFORM == VITE_STANDARD_PYTHON_MODE
             const backdrop = document.querySelector(".modal-backdrop") as HTMLElement | null;
             const graphicsEl = document.getElementById(getPEAGraphicsContainerDivId());
             if (!backdrop || !graphicsEl) {
@@ -535,6 +542,7 @@ export default defineComponent({
                 `${rect.left}px ${rect.top}px`, `0px ${rect.top}px`,
             ].join(", ");
             backdrop.style.clipPath = `polygon(${points})`;
+            // #v-endif
         },
 
         clearGraphicsColourPreview() {
--- a/src/components/ColourPickerDlg.vue
+++ b/src/components/ColourPickerDlg.vue
@@ -1,5 +1,5 @@
 <template>
-    <ModalDlg :dlgId="dlgId" :dlgTitle="dlgTitle" :okDisabled="okDisabled">
+    <ModalDlg dlgId="colourPickerDlg" :dlgTitle="$t('media.colourPickerTitle')" :okDisabled="okDisabled">
         <div class="ColourPickerDlg-content">
             <div class="ColourPickerDlg-view-area">
                 <div v-if="view === 'grid'" class="ColourPickerDlg-family-grid">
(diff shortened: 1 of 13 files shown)


src/components/ColourPickerDlg.vue
```diff
@@ -0,0 +1,820 @@
...
+    SHADE_ROW_SAT_COUNTS, GREY_SHADE_BANDS, familyHasHueChoice, midOfRange,
+} from "@/helpers/colour";
+
+const DEFAULT_HEX = "#ff0000";
+
+function clamp(x: number, lo: number, hi: number): number {
+    return Math.min(hi, Math.max(lo, x));
+}
+
+function hexColourDistance(a: string, b: string): number {
```
neilccbrown, line 132: Document this with a one line comment explaining it uses Pythagoras in RGB space.

src/components/ColourPickerDlg.vue
```diff
@@ -0,0 +1,820 @@
...
+    const ar = parseInt(a.substring(1, 3), 16), ag = parseInt(a.substring(3, 5), 16), ab = parseInt(a.substring(5, 7), 16);
+    const br = parseInt(b.substring(1, 3), 16), bg = parseInt(b.substring(3, 5), 16), bb = parseInt(b.substring(5, 7), 16);
+    return (ar - br) ** 2 + (ag - bg) ** 2 + (ab - bb) ** 2;
+}
+
+interface ShadeCell {
```
neilccbrown, line 141: Document this with 1-2 lines.  How come it has sat/light but not hue?  Presumably because it's dependent on the hue choice at top of dialog.

src/components/ColourPickerDlg.vue
```diff
@@ -0,0 +1,820 @@
...
+export default defineComponent({
+    name: "ColourPickerDlg",
+
+    components: {
+        ModalDlg,
+    },
+
+    props: {
+        dlgId: String,
+        dlgTitle: String,
```
neilccbrown: For this dialog, and for that matter for RecordImageDlg and RecordSoundDlg, why is this is a property?  I think it should just be inlined in all three dialogs unless there's a reason not to.

src/components/ColourPickerDlg.vue
```diff
@@ -0,0 +1,820 @@
...
+        // A hex code or recognised CSS colour name to seed the picker with; null/unparseable
+        // falls back to a default colour.
+        initialColour: {type: String as PropType<string | null>, default: null},
+    },
+
+    data: function () {
+        return {
+            view: "grid" as "grid" | "tinker" | "classic",
+            currentFamilyKey: null as string | null,
```
neilccbrown, line 167: Document this with 1 line.

src/components/ColourPickerDlg.vue
```diff
@@ -0,0 +1,820 @@
...
+        dlgId: String,
+        dlgTitle: String,
+        // A hex code or recognised CSS colour name to seed the picker with; null/unparseable
+        // falls back to a default colour.
+        initialColour: {type: String as PropType<string | null>, default: null},
+    },
+
+    data: function () {
+        return {
+            view: "grid" as "grid" | "tinker" | "classic",
```
neilccbrown, line 165: Document what the three views are.

src/components/ColourPickerDlg.vue
```diff
@@ -0,0 +1,820 @@
...
+
+        currentFamily(): ColourFamily | null {
+            return this.families.find((f) => f.key == this.currentFamilyKey) ?? null;
+        },
+
+        hasHue(): boolean {
+            return this.currentFamily != null && familyHasHueChoice(this.currentFamily);
+        },
+
+        hueChips(): {hue: number, hex: string, selected: boolean}[] {
```
neilccbrown, line 218: Document this with one line.

src/components/ColourPickerDlg.vue
```diff
@@ -0,0 +1,820 @@
...
+            const chipLight = family.chipLight ?? GOOD_CHIP_LIGHT;
+            const chips = [];
+            for (let i = 0; i < HUE_CHIP_COUNT; i++) {
+                const h = family.hue[0] + (family.hue[1] - family.hue[0]) * (i / (HUE_CHIP_COUNT - 1));
+                chips.push({hue: h, hex: hslToHex(h, chipSat, chipLight), selected: Math.abs(h - this.hue) < 0.01});
+            }
+            return chips;
+        },
+
+        shadeRows(): ShadeCell[][] {
```
neilccbrown, line 238: Document this with 1-3 lines

src/components/ColourPickerDlg.vue
```diff
@@ -0,0 +1,820 @@
...
+                    return rowCells;
+                });
+            }
+        },
+
+        currentHex(): string {
+            return this.view == "classic"
+                ? hsvToHex(this.classicHue, this.classicSat, this.classicVal)
+                : hslToHex(this.hue, this.sat, this.light);
+        },
```
neilccbrown, line 285: I wonder if it would be easier if the fine grained picker used HSL as well, instead of having it use a different model to the tile model?

src/components/ColourPickerDlg.vue
```diff
@@ -0,0 +1,820 @@
...
+        updateHexFromClassic() {
+            this.hasSelectedColour = true;
+            this.hexText = hsvToHex(this.classicHue, this.classicSat, this.classicVal);
+        },
+
+        // Called on user input into the always-visible hex text field. Mirrors the logic used when
+        // the dialog is first opened with an existing colour (see onShownModalDlg below): if the
+        // typed colour lands exactly on a tile swatch, jump to that tile selected in the tinker
+        // view; otherwise fall through to the fine-grained selector with the colour in place.
+        onHexTextEdited() {
```
neilccbrown: Can onShownModalDlg be changed to use this function, rather than mirroring the logic?

neilccbrown (commented):
Pointed out some changes to make.

PR #1052, opened 2026-08-11
Add colour literals: auto-detect/convert hex-colour strings to a colour swatch

neilccbrown:
## Summary
- Adds a third media-literal kind alongside images/sounds: a colour literal. Unlike those, its underlying Python code is just a plain quoted hex string (`"#aabbcc"`), reusing the existing generic `MediaSlot`/`SlotType.media` machinery (1-char-wide navigation, codegen, copy/paste, undo/redo) with `mediaType: "colour"`.
- Auto-converts a plain string to a colour literal when its content matches `#rrggbb` and the cursor leaves the string (typing-blur), and on loading/pasting a matching bare string.
- Ctrl+Shift+Y (existing colour-picker shortcut) now inserts/converts directly to a colour literal instead of a plain string, with the cursor landing in the adjacent field afterwards.
- Hovering a swatch shows a filled-rectangle preview with the hex as the header text and an Edit button (Preview/Download hidden, unlike images/sounds); Edit reopens the picker seeded with the current hex.
- Loading a `.spy`/`.py` program with a bare hex string renders it as a colour literal; saving round-trips the exact original text.

## Test plan
- [x] `tests/playwright/e2e/colour-literals.spec.ts` — typing/blur auto-conversion (case normalisation, both quote styles, cursor-still-inside guard, `f`/`r`/`b`-prefix guard, non-matching content, undo/redo), and load/save round-trip.
- [x] `tests/playwright/e2e/colour-picker.spec.ts` — updated existing insertion/replacement tests for the new `MediaSlot` shape, plus new cursor-placement and hover-popup/edit-round-trip coverage.
- [x] `tests/playwright/e2e/colour-literal-copy-paste.spec.ts` — copy puts the bare quoted hex text on the clipboard; paste recreates the swatch.
- [x] Full Playwright suite run twice on this branch's CI: only pre-existing, unrelated flaky tests (`scroll-into-view.spec.ts` etc.) remain red; all colour-literal-specific tests pass.
- [x] Cypress suite green.

### DIFF
--- a/docs/claude-colour-literals/PLAN.md
+++ b/docs/claude-colour-literals/PLAN.md
@@ -1,0 +1,268 @@
+# Colour literals — implementation plan
+
+Status: **not started**. Written up during planning (high effort), ready to execute
+(intended to be run at low effort, picking up straight from this file).
+
+## Goal
+
+Add a third kind of literal alongside the existing image/sound media literals:
+a colour literal. Unlike image/sound, it has no function-call underlying code —
+its underlying Python code is just a plain quoted string matching `#aabbcc`
+(6 hex digits). It should:
+
+- show a preview dialog (on hover) with a filled rectangle of the colour, plus
+  an Edit button that opens the existing colour-picker dialog (`ColourPickerDlg.vue`,
+  already wired up to Ctrl+Shift+Y)
+- auto-convert from a plain string to a colour literal when the user types a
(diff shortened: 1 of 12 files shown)


neilccbrown:
This is now ready for review.  It's not actually very big if you don't count the tests.  It now supports 8-digit or 6-digit hex codes.

PwtKCL (approved):

